PR #111, opened 2014-10-25
Fixes bug 1088743 - Make clear who may sign in

bugzPDX:
Make it clear user must sign in as a Mozillian not simply Persona.


peterbe:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1088743#c12

PR #156, opened 2014-12-25
Bug#1115451- Cache invalidation for deleting pictures

anurag90x:
When a picture is deleted the get_all_pictures key is removed from the cache but any event associated with the picture has the key as get_all_pictures+event.id in the cache. That's why the reference still persists. To rectify that, before rendering the page just check whether  the result from trying to get 'get_all_pictures' is the same as the result from the 'get_all_pictures+event.id' case. If it is, wonderful else something happened (this could be that the picture was deleted or a new picture was added or both happened, one after the other. In this case always go to the db and reset the cache values).

The test is only for the case mentioned in the bug.


anurag90x:
Have you had a chance to look at this one @peterbe ?

peterbe:
Ok. Now I've had a chance to look at it. I think the test you've added is good and can stay, I don't think we should rely on fetching from the cache twice or to check if there's 0 items. 

To me, a much better solution would be to use signals. See https://github.com/mozilla/airmozilla/blob/master/airmozilla/manage/models.py#L19

What we could do there is to "guess" what the other cache key might be if what's deleted is a `Picture` and if it has an event. Eg.

```python
#UNTESTED CODE!
@receiver(models.signals.post_save, sender=Event)
@receiver(models.signals.post_save, sender=Picture)
@receiver(models.signals.post_delete, sender=Picture)
def invalidate_picture_cache(sender, instance, **kwargs):
    cache_key = '_get_all_pictures'
    cache.delete(cache_key)
    if sender is Picture and instance.event_id:
        cache_key += str(instance.event_id)
        cache.delete(cache_key)
```

The reason for using `if .... and instance.event_id:` instead of `if .... and instance.event:` is that if you do that Django will unnecessarily make another SQL query (e.g. `select * from main_event where id=:id`) and then create instance from that and attach to `instance.event`. By just referencing `instance.event_id` you're just using the ID without turning the ID into an object with that ID. 
Does that make sense?

anurag90x:
screen shot 2015-01-06 at 12 22 50 am

Ok, so basically I put in the change from the comment and did the following:

1. Went to picture gallery (from management pane) and uploaded a random picture from my desktop.
2. Went to the test event through Event manager and saw above random pic.
3. Clicked delete on the pic (still on the same page)
4. Refreshed the page.
5. Saw the above screenshot.

Basically, deleting the _get_all_pictures key isn't sufficient and since this isn't from a screencap the if condition doesnt go through. But if this general picture case is unlikely/impossible to happen, then I guess it doesn't matter.

peterbe:
I see. It is the problem. And I can reproduce it too. 
This is getting silly. I think we should entirely reverse our thinking and remove all caching entirely. 

The query is really light and it's just getting too complicated to have different cache keys and mess around. 

What do you do think @anurag90x? Shall we just close this PR, close the bug and file a new bug for making things simpler?

Also, the picture gallery isn't so frequently used that the 60 seconds cache is doing anything super useful.

anurag90x:
@peterbe I agree. I think my fix takes care of this case but it just adds to the overall confusion when reading the code. When I looked at the code, I was wondering why there would be so much caching involved for getting the pictures , with the different cache keys and all. I think just having a key for all_pictures is enough. So if the request is for no particular event (/picturegallery), use previous cache result to display the data or if it's the first time then insert into cache, and if it's for an event (/picturegallery/?event=<event_id>), then just get the relevant results from the database without involving any additional caching.

I'd be happy to fix this in a new bug. :smile:

peterbe:
We've decided to not bother.

PR #157, opened 2014-12-27
Bug#1097055 - New page for editing questions

anurag90x:


airmozilla/manage/static/manage/css/survey.css
```diff
@@ -0,0 +1,16 @@
...
+    width: 80%;
+  }
+  textarea {
+    width: 100%;
+    height: 140px;
+  }
+  td.ordering {
+    vertical-align: middle;
+  }
+
```
peterbe: This file needs some cleaning up. Consistent spacing, same indentation and the last line needs to be a `\n`.

E.g.
```python
>>> open('airmozilla/manage/static/manage/css/survey.css').read()[-1]
' '
>>> open('airmozilla/surveys/static/surveys/css/survey.css').read()[-1]
'\n'
```
That's why GitHub puts a little red stop and return icon here. 

What editor are you using? You should be able to configure it to automatically fix that on save.

airmozilla/manage/templates/manage/survey_edit.html
```diff
@@ -9,19 +9,8 @@
   {% compress css %}
   <link href="{{ static('select2/select2.css') }}" rel="stylesheet" type="text/css">
   <link href="{{ static('select2/select2-bootstrap.css') }}" rel="stylesheet" type="text/css">
+  <link href="{{ static('manage/css/survey.css') }}" rel="stylesheet" type="text/css">
```
peterbe: I'm almost embarrassed by this but you need to do only `select2.css` and `select2-bootstrap.css` together in one block of `compress css`/`endcompress`.

The reason is that if you look for places where `select2-bootstrap.css` is used, throughout the project, you'll see that there are many places where it uses these two together. When the `compress css` thing works, it turns that into one file with a unique name. E.g. a3ed10efaa2.css and that gets reused throughout the management pages. So, if you make a combination of those two plus one more it gets a different unique name and the browser won't know it's already downloaded parts of it so you get excess downloaded. 

The solution would be to do this:
```html
{% compress css %}
<link href="{{ static('select2/select2.css') }}" rel="stylesheet" type="text/css">
<link href="{{ static('select2/select2-bootstrap.css') }}" rel="stylesheet" type="text/css">
{% endcompress %}
{% compress css %}
<link href="{{ static('manage/css/survey.css') }}" rel="stylesheet" type="text/css">
{% endcompress %}
```

airmozilla/manage/templates/manage/survey_edit.html
```diff
@@ -40,58 +29,24 @@
     {% include 'manage/_default_form.html' %}
 
     <h3>Questions</h3>
+    <a class="survey-links" href="{{ url('manage:survey_questions', survey.id) }}">Edit questions here</a>
```
peterbe: The class name `survey-links` is not a great name. A better name would be `survey-link`. It's the same with the class names in `models.py`. You call it `class Question` instead of `class Questions` even though you know there might/will be more than one of them.

peterbe: Also, instead of using CSS to make it a block element, wrap the `<a>` tag in something that *is* a block element like
```html
<p>
  <a class="survey-link" ...>
</p>
```

airmozilla/manage/templates/manage/survey_edit.html
```diff
@@ -40,58 +29,24 @@
...
-            </form>
-            <form method="post" class="confirm"
-                  action="{{ url('manage:survey_question_delete', survey.id, question.id) }}">
-              {{ csrf() }}
-              <button class="btn btn-danger btn-sm" type="submit">
-                <i class="glyphicon glyphicon-trash"></i>
-                Delete
-              </button>
-            </form>
+            <textarea readonly name="question">{{ json_print(question.question, indent=2) }}</textarea>
```
peterbe: If you can't edit it, there's no point this being a `textarea` tag. Use a `<pre>` tag or a plain `<div>` tag and set `overflow:auto` on it.

airmozilla/manage/templates/manage/survey_edit_questions.html
```diff
@@ -0,0 +1,82 @@
...
+    Questions for "{{survey.name}}"
+{% endblock %}
+
+{% block site_css %}
+  {{ super() }}
+  {% compress css %}
+  <link href="{{ static('select2/select2.css') }}" rel="stylesheet" type="text/css">
+  <link href="{{ static('select2/select2-bootstrap.css') }}" rel="stylesheet" type="text/css">
+  <link href="{{ static('manage/css/survey.css') }}" rel="stylesheet" type="text/css">
+  {% endcompress %}
```
peterbe: Same point here as was made above.

airmozilla/manage/templates/manage/survey_edit_questions.html
```diff
@@ -0,0 +1,82 @@
...
+
+{% block site_js %}
+  {{ super() }}
+  <script src="{{ static('select2/select2.min.js') }}"></script>
+  {% compress js %}
+    <script src="{{ static('manage/js/survey-edit.js') }}"></script>
+  {% endcompress %}
+{% endblock %}
+
+{% block content_class %}col-md-6{% endblock %}
```
peterbe: If you remove this line, I think you get a full width page and then the textarea boxes get a bit more breathing room.

airmozilla/manage/templates/manage/survey_edit_questions.html
```diff
@@ -0,0 +1,82 @@
...
+
+    <form method="post" action="{{ url('manage:survey_question_new', survey.id) }}">
+      {{ csrf() }}
+      <button class="btn btn-primary">
+        <i class="glyphicon glyphicon-plus-sign"></i>
+        New question
+      </button>
+    </form>
+
+    <a class="survey-links" href="{{ url('manage:survey_edit', survey.id)}}">Back to editing "{{survey.name}}"</a>
```
peterbe: Same thing here. You'll see other examples of this, [like this](https://github.com/mozilla/airmozilla/blob/master/airmozilla/manage/templates/manage/event_discussion.html#L50-L52)

airmozilla/manage/urls.py
```diff
@@ -76,6 +76,8 @@
     url(r'^surveys/(?P<id>\d+)/$', views.survey_edit, name='survey_edit'),
     url(r'^surveys/(?P<id>\d+)/delete/$', views.survey_delete,
         name='survey_delete'),
+    url(r'^surveys/(?P<id>\d+)/questions/$', views.survey_questions,
+        name='survey_questions'),
```
peterbe: Great!

airmozilla/manage/templates/manage/survey_edit_questions.html
```diff
@@ -0,0 +1,82 @@
...
+  {{ super() }}
+  <script src="{{ static('select2/select2.min.js') }}"></script>
+  {% compress js %}
+    <script src="{{ static('manage/js/survey-edit.js') }}"></script>
+  {% endcompress %}
+{% endblock %}
+
+{% block content_class %}col-md-6{% endblock %}
+
+{%block manage_content %}
```
peterbe: spacing

airmozilla/manage/templates/manage/survey_edit_questions.html
peterbe: Also, there's a pattern of using `&larr; Return ...` in these little links.

peterbe: Please using `"{{ survey.name }}"`

peterbe:
It's looking great! Some nits to work on. Looking forward to an update. 

By the way, when you add more commits you don't need to bother rebasing them into 1 commit. Not until you have an official r+.

airmozilla/manage/templates/manage/survey_edit_questions.html
anurag90x: Hey @peterbe. What do you mean by the please using {{survey.name}} comment? :confused:

airmozilla/manage/templates/manage/survey_edit_questions.html
anurag90x: hmm, as in the spacing between the lines? :question:

airmozilla/manage/templates/manage/survey_edit_questions.html
peterbe: `"{{[SPACE]survey.name[SPACE]}}"`

airmozilla/manage/templates/manage/survey_edit_questions.html
peterbe: `{% block manage_content %}` instead of `{%block manage_content %}`.

I know it sounds like an unimportant detail. And it is. And it's not. Quality is important because it makes projects easier to maintain. 
Have you heard of the [Broken Window Theory](https://en.wikipedia.org/wiki/Broken_windows_theory)
Basically, if an abandoned building has broken windows, people think that it's so abandoned that anybody can spray graffiti or demolish it or do whatever they like. Just because it has broken windows. But if the windows are all fixed people leave it alone and don't make it worse. 

The same goes with code. If the code is neat, structured and consistent people who come and work on it will continue to uphold that tradition and will continue to follow that line. 
If the code is sloppy and different in different places it sets your mind that it's not important what you write so the quality slowly drops. 

And consistency is important because it makes it easier to read the code because you're mind doesn't get hung up on odd differences. 

Sorry for the long essay. I hope it helps. I hope the airmozilla code is going to last for a long time and continue to be maintained for many years.

anurag90x: No problem! I just missed it. So I was looking though the diff for 5 mins really intently and then asked :smile:

peterbe:
As soon as I've fixed master, I'll let you know and you can rebase so you get that fix into this PR.

peterbe:
Now that https://github.com/mozilla/airmozilla/commit/90f29ea7b0ccf8884c9a0aab9daa93c050275953 has landed, rebase up this PR and it should pass tests.

anurag90x:
The div isn't of class question so div.question doesn't seem to work. But .question div does, so I did that instead. :smile:

peterbe:
This is looking really good. There's just one last problem. On the edit_survey.html page, there is now no longer a need for table. Also, instead of `<div class="question">{{ json_print(...) }}</div>` change that to `<pre class="question">{{ json_print(...) }}</pre>`.

So, perhaps instead of using a table (with the now unnecessary `<th>JSON</th>` stuff, just loop over the questions and do something like:
```html
{% for question in questions %}
    <div>
            <pre name="question">{{ json_print(question.question, indent=2) }}</pre>
    </div>
{% endfor %}
```

peterbe:
You rock!

PR #167, opened 2015-01-09
Bug#1117886 - Button to delete all associated pictures for an event

anurag90x:


airmozilla/manage/static/manage/js/picturegallery.js
```diff
@@ -240,6 +248,26 @@ app.controller('PictureGalleryController', ['$scope', '$http',
...
+                'csrfmiddlewaretoken': csrf,
+                'event': $scope.current_event
+            };
+            $http({
+                method: 'POST',
+                url: url,
+                headers: {'Content-Type': 'application/x-www-form-urlencoded'},
+                data: serializeObject(data)
+            }).success(function(data) {
+                $scope.pictures = [];
```
anurag90x: I don't think $scope.pictures = [] is the right approach because if there is a picture in this view which does not have an event_id (just some random picture) then it would appear as if it's been deleted but if you refresh the page , that of course would not be the case. I think maybe we should pass in the picture.event_id info to the context and only remove those pictures from the view that have a picture.event_id.

airmozilla/manage/views.py
```diff
@@ -3687,6 +3688,18 @@ def picture_delete(request, id):
 
 
 @staff_required
+@permission_required('main.delete_picture')
+@transaction.commit_on_success
+@json_view
+def picture_delete_all(request):
+    event = get_object_or_404(Event, id=request.POST.get('event'))
+    Picture.objects.filter(event=event).delete()
+    event.picture = None
```
peterbe: This is not necessarily true. 

Suppose an event has a picture associated with it from the general pick of pictures in the gallery. Then, it shouldn't reset the `event.picture`. I.e. we need something like this:
```pyhon
pictures = Picture.objects.filter(event=event)
if event.picture and event.picture in pictures:
    assert event.placeholder_img
    event.picture = None
    event.save()
pictures.delete()
```

See what I mean? Also, it would be great if you could add a dedicated test that tests both paths. One test where the event has a picture from those belonging to the event and one where the event has a picture but it's a general picture not belonging to any event.

Sorry. I understand that this "cross foreign key" situation is confusing.

airmozilla/manage/views.py
```diff
@@ -3687,6 +3688,18 @@ def picture_delete(request, id):
 
 
 @staff_required
+@permission_required('main.delete_picture')
+@transaction.commit_on_success
+@json_view
+def picture_delete_all(request):
+    event = get_object_or_404(Event, id=request.POST.get('event'))
```
peterbe: Can you change this to:
```python
event = get_object_or_404(Event, id=request.POST['event'])
```
The reason is that you only use `.get()` if it's "ok" to not be there. It then automatically becomes `None` so you could potentially do: `event = get_object_or_404(Event, id=None)`. It would fail but it would be cryptic why or how it failed.

airmozilla/manage/urls.py
```diff
@@ -209,6 +209,9 @@
     url(r'^picturegallery/(?P<id>\d+)/delete/$',
         views.picture_delete,
         name='picture_delete'),
+    url(r'^picturegallery/delete_all/$',
```
peterbe: Actually, can we change this to:
```
url(r'^picturegallery/(?P<id>\d+)/delete-all/$'
```
And change the view function to be:
```python
def picture_delete_all(request, id):
   event = get_object_or_404(Event, id=id)
```
That way, the system automatically checks that you don't accidentally call `POST /picturegallery/delete_all/` without a `?event=123`.

airmozilla/manage/views.py
```diff
@@ -3581,6 +3581,7 @@ def picturegallery_data(request):
     context['urls'] = {
         'manage:picture_edit': reverse('manage:picture_edit', args=('0',)),
         'manage:picture_delete': reverse('manage:picture_delete', args=('0',)),
+        'manage:picture_delete_all': reverse('manage:picture_delete_all'),
```
peterbe: With my comment earlier this'll need to become `reverse('manage:picture_delete_all', args=('0',))`

airmozilla/manage/views.py
```diff
@@ -3687,6 +3688,18 @@ def picture_delete(request, id):
 
 
 @staff_required
```
peterbe: Add also, `@require_POST` to this view.

airmozilla/manage/templates/manage/picturegallery.html
```diff
@@ -16,56 +16,81 @@
...
+
+    {% else %}
+
+     <div class="alert alert-warning alert-dismissible" role="alert">
+      <button type="button" class="close" data-dismiss="alert">
+        <span aria-hidden="true">&times;</span><span class="sr-only">Close</span>
+      </button>
+      <p>
+        Unable to delete all pictures associated with this event because the event
+        relies on one of them and has no fallback placeholder image.
```
peterbe: excellent!

airmozilla/manage/static/manage/js/picturegallery.js
peterbe: You're right. That's not going to be enough. We'll need two things
1. Something like this:
```javascript
$scope.pictures = $scope.pictures.filter(function(picture) {
    return picture.event !== CURRENT_EVENT;
});
```

And when we change the [view function](https://github.com/anurag90x/airmozilla/blob/Bug%231117886/airmozilla/manage/views.py#L3621-L3648) to also include this `event` key. 

Does that make sense?

vendor
```diff
@@ -1 +1 @@
-Subproject commit 2a0c7ade84641cd9de47d9aa8120caad97c4ef79
+Subproject commit ec29c24cf607f8cc35c907b0d0d0d04fe9d2b6b8
```
peterbe: Submodules are confusing. What your PR now says is to downgrade to ec29c24cf607f8cc35c907b0d0d0d04fe9d2b6b8 which was the previous pointer for the submodule. 
I think if you do `cd vendor; git pull origin master; cd ..; git commit -a -m "current vendor submodule"` it should correct itself. If not, ping me on IRC and I'll help.

airmozilla/manage/static/manage/js/picturegallery.js
anurag90x: Yup makes sense.

airmozilla/manage/views.py
anurag90x: Haha yeah I get it. I didn't think of that possibility when I wrote it.

airmozilla/manage/templates/manage/picturegallery.html
```diff
@@ -16,56 +16,81 @@
...
+    </h4>
+
+    {% if event.placeholder_img %}
+
+      <p>
+        <a class="btn btn-danger"
+           ng-show="!deleteAllConfirmation"
+           ng-click="toggleDeleteAllConfirmation()">
+          <i class="glyphicon glyphicon-minus-sign"></i>
+          Remove all pictures for {{ event.title }}
```
peterbe: This could make the button really really long. 
screenshot 2015-01-09 09 04 09

Instead, let's change it to something like "Remove all pictures made for this event". What do you think?

anurag90x: Wow that is super long. Yup, I agree.

airmozilla/manage/views.py
```diff
@@ -3643,7 +3647,8 @@ def _get_all_pictures(event=None):
             'height': picture.height,
             'size': picture.size,
             'created': picture.created.isoformat(),
-            'events': event_map[picture.id]
+            'events': event_map[picture.id],
+            'event': picture.event
```
peterbe: Change this to `'event': picture.event_id` instead. That way it doesn't make another SQL query just to materialize that into an ORM object.

peterbe:
Works like a charm! Would you mind rebasing this into one nice commit?

peterbe:
R+

airmozilla/manage/views.py
anurag90x: Hmm  'event': picture.event_id is giving me a key error

anurag90x: Nvm. Fixed.

PR #168, opened 2015-01-09
Fixes bug 1097180

lcamacho:


airmozilla/manage/archiver.py
```diff
@@ -23,10 +23,13 @@ def auto_archive():
                 template__name__contains='Vid.ly')
     )
     for event in events:
-        archive(event, swallow_email_exceptions=True)
+        archive(event, swallow_email_exceptions=True, verbose=verbose)
 
+    if verbose:  # pragma: no cover
+        print "No pending events"
```
peterbe: This would now happen even if there's nothing in the loop.

airmozilla/manage/cron.py
```diff
@@ -34,7 +34,7 @@ def cron_ping():
 @cronjobs.register
 @capture
 def auto_archive():
-    archiver.auto_archive()
+    archiver.auto_archive(verbose=True)
```
peterbe: perfect!

airmozilla/manage/archiver.py
```diff
@@ -44,6 +52,9 @@ def archive(event, swallow_email_exceptions=False):
                     event,
                     tag,
                 )
+                if verbose:  # pragma: no cover
+                    print 'Unable to find Vid.ly video with tag %s \
```
peterbe: When you have a multiline string like this use the brackets instead:
```python
print (
    'Unable to find Vid.ly video with tag %s for this event "%s"' % (
        tag,
        event
    )
)
```

airmozilla/manage/archiver.py
```diff
@@ -63,6 +74,9 @@ def archive(event, swallow_email_exceptions=False):
                     event,
                     tag,
                 )
+                if verbose:  # pragma: no cover
+                    print 'Unable to archive pending event "%s" \
+                            with tag %s' % (event, tag)
```
peterbe: ditto

airmozilla/manage/archiver.py
```diff
@@ -80,6 +94,9 @@ def archive(event, swallow_email_exceptions=False):
         event.status = Event.STATUS_SCHEDULED
         event.save()
 
+        if verbose:  # pragma: no cover
+            print 'Event "%s" archived!' % event
```
peterbe: Really useful!

peterbe:
Couple of nits to fix. Ping me when you've pushed another update.

peterbe:
...so I don't miss the notification and leave you waiting too long.

airmozilla/manage/archiver.py
```diff
@@ -23,10 +23,13 @@ def auto_archive():
                 template__name__contains='Vid.ly')
     )
     for event in events:
-        archive(event, swallow_email_exceptions=True)
+        archive(event, swallow_email_exceptions=True, verbose=verbose)
 
+        if event == events[-1] and verbose:  # pragma: no cover
```
peterbe: Did you try this? I don't think it works. 

`events` is not a list. It's a QuerySet object that you can iterate on like a list. 
So something like this:
```python
any = False
for event in events:
    archive(event, ...)
    any = True
if any:
    ...
```

peterbe:
Excellent! Thank you!

Do you need another bug to work on? I've got plenty!

PR #180, opened 2015-01-22
Bug 1097967 - Add Regions Table to DB

bugzPDX:
Replacing with a fresh copy to fix runserver errors

Bug 1097967 - Updates for review - Add Regions Table

Bug 1097967 - Add migration for Add Regions Table

Bug 1097967 - Edits to region_edit.html

Ticket 1097967 - Updates for Add Region Table to DB

Removing deprecated vendor-local directory


airmozilla/main/models.py
```diff
@@ -20,7 +20,7 @@
 
 
 def _get_now():
-    return timezone.now()
+    return datetime.datetime.utcnow().replace(tzinfo=utc)
```
peterbe: This change, and the change above must have been a little mistake when you rebased or something. The correct code should be:
```python
from django.utils import timezone
...
def _get_now():
    return timezone.now()
```

airmozilla/main/models.py
```diff
@@ -189,11 +189,36 @@ def __unicode__(self):
...
+    def __unicode__(self):
+        return self.name
+
+
 class Location(models.Model):
     """Venue/location of a video/stream/presentation/etc."""
     name = models.CharField(max_length=300)
     timezone = models.CharField(max_length=250)
     is_active = models.BooleanField(default=True)
+    regions = models.ManyToManyField(Region, blank=True, null=True,)
```
peterbe: You can remove the `blank=True, null=True,` part.

airmozilla/main/models.py
```diff
@@ -189,11 +189,36 @@ def __unicode__(self):
...
+        ordering = ['name']
+
+    def __unicode__(self):
+        return self.name
+
+
+class Region(models.Model):
+    """Region of a video/stream/presentation/etc."""
+    name = models.CharField(max_length=300)
+    is_active = models.BooleanField(default=True)
```
peterbe: Duplicate class.

airmozilla/manage/templates/manage/locations.html
```diff
@@ -25,6 +26,9 @@
           <span class="label label-default" title="Inactive location">Inactive</span>
           {% endif %}
         </td>
+        <td>{% for region in location.regions.all() %}
+        {{ region.name }}
```
peterbe: To add a comma between each you can do this:
```
{{ region.name }}{% if not loop.last %}, {% endif %}
```

airmozilla/manage/views/regions.py
```diff
@@ -0,0 +1,75 @@
...
+        form = forms.RegionEditForm()
+    return render(request, 'manage/region_new.html', {'form': form})
+
+
+@staff_required
+@permission_required('main.delete_region')
+@transaction.commit_on_success
+def region_remove(request, id):
+    region = get_object_or_404(Region, id=id)
+    if request.method == 'POST':
```
peterbe: Remove this line and add a new decorator to this view function called `@require_POST` You can find examples of it in other views.

airmozilla/manage/templates/manage/locations.html
```diff
@@ -11,6 +11,7 @@
     <thead>
       <tr>
         <th>Name</th>
+        <th>Region</th>
```
peterbe: Maybe change this to "Region(s)".

airmozilla/manage/templates/manage/regions.html
```diff
@@ -0,0 +1,51 @@
+{% extends 'manage/manage_base.html' %}
```
peterbe: I made a small PR towards your PR on this https://github.com/bugzPDX/airmozilla/pull/1

airmozilla/main/models.py
```diff
@@ -20,7 +20,7 @@
 
 
 def _get_now():
-    return timezone.now()
+    return timezone.now
```
peterbe: missing `()`

airmozilla/main/models.py
```diff
@@ -9,7 +9,7 @@
 from django.db import models
 from django.db.models import Q
 from django.dispatch import receiver
-from django.utils import timezone
+from django.utils.timezone import utc
```
peterbe: This diff should be removed. It should just be `from django.utils import timezone`

airmozilla/main/models.py
```diff
@@ -189,11 +189,24 @@ def __unicode__(self):
...
+    def __unicode__(self):
+        return self.name
+
+
 class Location(models.Model):
     """Venue/location of a video/stream/presentation/etc."""
     name = models.CharField(max_length=300)
     timezone = models.CharField(max_length=250)
     is_active = models.BooleanField(default=True)
+    regions = models.ManyToManyField(Region,)
```
peterbe: No extra comma needed.

bugzPDX:
Added tests @peterbe for you to look over.

airmozilla/manage/tests/views/test_locations.py
```diff
@@ -19,6 +20,23 @@ def test_locations(self):
...
+        """Location management pages with regions return successfully."""
+        response = self.client.get(reverse('manage:locations'))
+        eq_(response.status_code, 200)
+
+        location = Location.objects.create(name='SomeLocation')
+        region = Region.objects.create(name='SomeRegion')
+        response = self.client.get(reverse('manage:locations'))
+        eq_(response.status_code, 200)
+        eq_(location.name, 'SomeLocation')
+        eq_(region.name, 'SomeRegion')
```
peterbe: These two should be 
```python
ok_('SomeLocation' in response.content)
ok_('SomeRegion' in response.content)
```

airmozilla/manage/tests/views/test_regions.py
```diff
@@ -0,0 +1,67 @@
...
+class TestRegions(ManageTestCase):
+    def test_regions(self):
+        """Region management pages return successfully."""
+        response = self.client.get(reverse('manage:regions'))
+        eq_(response.status_code, 200)
+
+        region = Region.objects.create(name='New Region')
+        response = self.client.get(reverse('manage:regions'))
+        eq_(response.status_code, 200)
+        eq_(region.name, 'New Region')
```
peterbe: ditto

peterbe: Add another test to list 1 inactive region. E.g. 
```python
def test_regions_inactive(self):
    Region.objects.create(name='New Region', is_active=False)
    response = self.client.get(reverse('manage:regions'))
    eq_(response.status_code, 200)
    ok_("Not active!" in response.content)
```
(note: I can't remember what the little label says next to the region name if it's not active).

airmozilla/manage/tests/views/test_regions.py
```diff
@@ -0,0 +1,67 @@
...
+from .base import ManageTestCase
+
+
+class TestRegions(ManageTestCase):
+    def test_regions(self):
+        """Region management pages return successfully."""
+        response = self.client.get(reverse('manage:regions'))
+        eq_(response.status_code, 200)
+
+        region = Region.objects.create(name='New Region')
```
peterbe: You'll find you won't need this `region = ` variable.

airmozilla/manage/tests/views/test_regions.py
```diff
@@ -0,0 +1,67 @@
...
+        eq_(response.status_code, 200)
+
+        response_ok = self.client.post(url, {
+            'name': 'testing',
+        })
+        self.assertRedirects(response_ok, reverse('manage:regions'))
+        region = Region.objects.get(name='testing')
+        response_fail = self.client.post(url)
+        eq_(response_fail.status_code, 200)
+        eq_(region.name, 'testing')
```
peterbe: this line is not necessary.

airmozilla/manage/tests/views/test_regions.py
```diff
@@ -0,0 +1,67 @@
...
+        region = Region.objects.create(
+            name="Something"
+        )
+        self._delete_test(
+            region,
+            'manage:region_remove',
+            'manage:regions'
+        )
+
+        region = Region.objects.get(id=1)
```
peterbe: You're referring to something in the fixture here. Then I would suggest you fetch it by its name instead. That's easier to read.

airmozilla/manage/tests/views/test_regions.py
```diff
@@ -0,0 +1,67 @@
...
+         with null regions."""
+        region = Region.objects.create(
+            name="Something"
+        )
+        self._delete_test(
+            region,
+            'manage:region_remove',
+            'manage:regions'
+        )
+
```
peterbe: `self._delete_test` deleted it. 
Then you should do something like 
```python
# it's been deleted
assert not Region.objects.filter(name="Something")
```

airmozilla/manage/tests/views/test_regions.py
```diff
@@ -0,0 +1,67 @@
...
+        region = Region.objects.get(id=1)
+        url = reverse('manage:region_remove', args=(region.id,))
+        response = self.client.post(url)
+        eq_(response.status_code, 302)
```
peterbe: These 4 lines is already what the `self._delete_test()` does.

airmozilla/manage/tests/views/test_regions.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def test_region_edit(self):
+        """Test region editor"""
+        url = reverse('manage:region_edit', kwargs={'id': 1})
+        response = self.client.get(url)
+        region = Region.objects.get(id=1)
+        eq_(response.status_code, 200)
+        ok_(region.name != 'North America')
```
peterbe: So, ID 1 should correspond to a region called "South America" right?
Then you should expect that to appear if you do a GET on the edit view. 
So maybe the last line should be `ok_('South America' in response.content)`

airmozilla/manage/tests/views/test_regions.py
```diff
@@ -0,0 +1,67 @@
...
+
+        response_ok = self.client.post(url, {
+            'name': 'North America',
+        })
+        self.assertRedirects(response_ok, reverse('manage:regions'))
+        region = Region.objects.get(id=1)
+        eq_(region.name, 'North America')
```
peterbe: You can replace these two lines with something that has the test "built in".
E.g. "ok_(Region.objects.get(name="North America")`

peterbe:
I have a bigger general comment out fixtures. It's basically something we want to avoid. 
Fixtures should ideally be slim. 

Basically, what I think you should do is this:

1. Revert the change you made to the fixture json file
2. In `class TestRegions(ManageTestCase):` add something like this:
```python
def setUp(self):
    super(TestRegions, self).setUp()
    Regions.objects.create(name='South America')
```
that way it's available to all tests under that class. 
3. In the one test in `test_locations.py` you can manually add a region.

bugzPDX:
More updates for your review @peterbe

peterbe:
r+ Yay!

Now, would you please rebase that into one single commit and push back into the PR and I'll merge it.

bugzPDX:
I sure will! Whew :-)

PR #183, opened 2015-01-23
Wide layout

craigcook:
Bit of a work-in-progress, I suspect there may be some more templates to touch.

There may be a better way of doing this, but I've added a block for 'body_class' in the base template and any pages that should have a sidebar get the class 'with-sidebar'. Other pages default to an empty class and the wide layout, so you really only have to touch templates that need a sidebar. See home.html for example.

### DIFF
--- a/airmozilla/base/templates/base.html
+++ b/airmozilla/base/templates/base.html
@@ -20,7 +20,7 @@
     </script>
     {% endif %}
   </head>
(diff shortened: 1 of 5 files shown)


peterbe:
Rebased and merged in https://github.com/mozilla/airmozilla/commit/26d03cc982b9d4ed45260009a258b63cd61c1e62

PR #186, opened 2015-01-30
Fixes bug 1126007 - Show weekend on Calendar page

lcamacho:


lcamacho:
@peterbe I'm not sure of the place where I put jstz script, but no idea where else I should put It.

peterbe:
Great start! I think the right "pattern" to follow is to put the file in `airmozilla/main/static/main/jstz/jstz-1.0.4.min.js`. I.e. a directory called `jstz`. That's what we do with the fullcalendar javascript.

airmozilla/main/templates/main/calendar.html
```diff
@@ -19,6 +19,7 @@
   {{ super() }}
   {% compress js %}
   <script src="{{ static('main/fullcalendar/fullcalendar.min.js') }}"></script>
+  <script src="{{ static('main/js/jstz-1.0.4.min.js') }}"></script>
```
peterbe: So, this should then be
```html
<script src="{{ static('main/jstz/jstz-1.0.4.min.js') }}"></script>
```

airmozilla/main/static/main/js/calendar.js
```diff
@@ -1,31 +1,34 @@
...
+            header: {
+                left: 'prev,next today',
+                center: 'title',
+                right: 'month,agendaWeek,agendaDay'
+            }
+        });
+    };
+    startsOnMonday.on('change', function(){
+        container.fullCalendar('destroy');
+        initCalendar();
```
peterbe: Instead of re-creating the whole thing, can you see if it can be done with http://fullcalendar.io/docs/display/render/

airmozilla/main/static/main/js/calendar.js
```diff
@@ -1,31 +1,34 @@
...
-            center: 'title',
-            right: 'month,agendaWeek,agendaDay'
-        }
+    var startsOnMonday = $('input[type="checkbox"]');
+    // if user isn't in American TZ then Monday is first day of week
+    var tz = jstz.determine().name();
+    if (!tz.match('^America/')) {
+        startsOnMonday.prop('checked', true);
+    }
+    initCalendar = function() {
```
peterbe: lacks a `var `

airmozilla/main/static/main/js/calendar.js
```diff
@@ -1,31 +1,34 @@
...
-        allDaySlot: false,
-        defaultTimedEventDuration: '01:00:00',
-        timeFormat: 'h:mm a',
-        header: {
-            left: 'prev,next today',
-            center: 'title',
-            right: 'month,agendaWeek,agendaDay'
-        }
+    var startsOnMonday = $('input[type="checkbox"]');
+    // if user isn't in American TZ then Monday is first day of week
```
peterbe: So this is a hack and we should explain that. We had a lengthy discussion about this on IRC. 
The conclusion was basically that we're "biased" against United States visitors who are in a "America/Something" time zone. That's because such a large majority of our visitors are from the US so we're given them the pleasure of the default matching their preference.

airmozilla/main/templates/main/calendar.html
```diff
@@ -29,6 +30,8 @@
   <h2 class="section-title">Calendar</h2>
   <div id="calendar" data-events-url="{{ url('main:calendar_data') }}"></div>
 
+  <input type="checkbox"> Week starts on Monday.</input>
```
peterbe: That's not how you do a checkbox input. It needs to have an id and a label so that if you click the label in toggles the checkbox. Also, an input tag is not a block element so it needs to be wrapped in one. Try something like this:
```html
<p>
  <input type="checkbox" id="id_startsonmonday">
 <label for="id_startsonmonday">Week starts on a Monday</label>
</p>
```

The `label` tag might have some CSS inherited that you might need to wrangle.

peterbe:
Sorry if I'm making more work for you but there's one big thing missing; memory.

A user shouldn't have to check that checkbox twice. We have some options:

1) Using a server-side HTTPS cookie (requires an AJAX post after you've toggled)
2) Using localStorage (synchronous)
3) Storing this preference in the [UserProfile](https://github.com/mozilla/airmozilla/blob/master/airmozilla/main/models.py#L31) (requires an AJAX post)
4) Using IndexedDB (wrapped by [localForage](https://github.com/mozilla/localforage/)) which is asyncronous

I'm personally in favor of option 2. That's after all what that was built for. You could simply do something like this:
```javascript
...
if (localStorage.getItem('startsonamonday')) {
    startsOnMonday.prop('checked', true);
}
initCalendar();
...
```

airmozilla/main/static/main/js/calendar.js
lcamacho: Unfortunately render doesn't take any parameter to modify the original
calendar, I tried other ways without success.

Saludos,
Leonard Camacho
Ingeniero en Computación

2015-02-04 17:35 GMT+00:00 Peter Bengtsson <notifications@github.com>:

peterbe: Event if you do it right after `container.fullCalendar('destroy');` ?

peterbe: If that's the case, just ignore my comment :)

lcamacho: The thing is that render doesn't take any parameter to be able to modify the original calendar, so there's no way to tell 'hey now you first day is monday, re render', and after destroy theres nothing to ask for render but instead I have to create It again.

lcamacho:
# 2 looks good, we need to check if true or false what is saved in that
item, but are we going to mantain the detection if not in 'Americas/' TZ
then Monday right? as a first approach to show calendar.



2015-02-04 13:24 GMT-04:30 Peter Bengtsson <notifications@github.com>:

airmozilla/main/static/main/js/calendar.js
peterbe: But can you do something like this:
```javascript
// change and re-render
container.fullCalendar({
    firstDay: toggleFirstDay
});
//container.fullCalendar('destroy');
container.fullCalendar('render');
```

peterbe:
I don't think you should store "false". If someone wants it on, you store something. E.g. `localStorage.setItem('startsonmonday', 1)` and if someone un-checks that box you do a `localStorage.deleteItem('startsonmonday')`.

airmozilla/main/static/main/js/calendar.js
lcamacho: I just tried, didn't work

peterbe: So be it then. Let's go for the full re-creation.

lcamacho:
What happens when people who is outside of America tz comes and unmark the checkbox, so next time the visit starts on sunday like they want, I think for this situation why should keep true or false (1 or 0) on that item, when theres no item why try guessing detecting time zone ans setting the item for first time.

peterbe:
> What happens when people who is outside of America tz comes and unmark the checkbox,

You are right! That's a great point. You're right, we should not use `localStorage.deleteItem()`. We should basically only allow the default behaviour if `localStorage.getItem('startsonmonday')` returns `null`. It's fine for it to return `true` or `false`.

lcamacho:
Last update includes that comment

airmozilla/main/static/main/js/calendar.js
```diff
@@ -1,31 +1,47 @@
...
+    var startsOnMonday = $('#startsOnMonday');
+    var storage_startsOnMonday = localStorage.getItem('startsOnMonday');
+    if (storage_startsOnMonday === null) {
+        // this is a hack to try to determine if week starts on Sunday or Monday
+        // right now if user is in America's TZ the week would start on Sunday
+        // if user is in another TZ It would start on Monday
+        var tz = jstz.determine().name();
+        if (!tz.match('^America/')) {
+            startsOnMonday.prop('checked', true);
+            localStorage.setItem('startsOnMonday', 1);
```
peterbe: No, we should only use `setItem` if they have genuinely made a choice.

airmozilla/main/static/main/js/calendar.js
```diff
@@ -1,31 +1,47 @@
...
+            timeFormat: 'h:mm a',
+            header: {
+                left: 'prev,next today',
+                center: 'title',
+                right: 'month,agendaWeek,agendaDay'
+            }
+        });
+    };
+    startsOnMonday.on('change', function(){
+        localStorage.setItem('startsOnMonday', startsOnMonday.prop('checked')? 1 : 0);
```
peterbe: excellent!

airmozilla/main/static/main/js/calendar.js
```diff
@@ -1,31 +1,44 @@
...
+    if (storage_startsOnMonday === null) {
+        // this is a hack to try to determine if week starts on Sunday or Monday
+        // right now if user is in America's TZ the week would start on Sunday
+        // if user is in another TZ It would start on Monday
+        var tz = jstz.determine().name();
+        if (!tz.match('^America/')) {
+            startsOnMonday.prop('checked', true);
         }
+    } else {
+        if (storage_startsOnMonday == 1) {
```
peterbe: This is an ugly side of javascript. It's actually doing `if ("1" == 1) {` and it becomes `true`. 
It works to our advantage in this case. :)

airmozilla/main/static/main/js/calendar.js
```diff
@@ -1,31 +1,44 @@
...
-        firstDay: firstDay,
-        allDaySlot: false,
-        defaultTimedEventDuration: '01:00:00',
-        timeFormat: 'h:mm a',
-        header: {
-            left: 'prev,next today',
-            center: 'title',
-            right: 'month,agendaWeek,agendaDay'
+    var startsOnMonday = $('#startsOnMonday');
+    var storage_startsOnMonday = localStorage.getItem('startsOnMonday');
```
peterbe: You can just call this `storage`. The name doesn't have to be perfect or unique because this code is already in a closure.

airmozilla/main/static/main/js/calendar.js
```diff
@@ -1,31 +1,44 @@
...
-    var currentTimeZoneOffsetInHours = x.getTimezoneOffset() / 60;
-    if (currentTimeZoneOffsetInHours <= -7) {
-        // sufficiently far east that some Friday events might actually
-        // start on a Saturday
-        weekends = true;
-        firstDay = 1;  // far-east people prefer Monday
-    } else {
-        weekends = false;
-    }
+    var weekends = true;
```
peterbe: You can remove this now and just set:
```javascript
weekends: true,
```
further down.

airmozilla/main/static/main/css/calendar.css
```diff
@@ -29,4 +29,8 @@ button {
 
 button:hover {
     box-shadow: none;
+}
+
+label {
+  display: inline;
 }
```
peterbe: For some reason your editing isn't adding one last empty line.

peterbe:
r+ with nits.

Yay! I'm really excited about this. Thank you for all your hard work. 

Now, one thing I've noticed personally is that I often prefer the weekly or daily view myself. I almost never look at the whole month. However, I think the month view should still be the default because it's clear what you're looking at then. But I think we should use localStorage to record which view they prefer and use that as the new default view when you load the page (if you have something in localStorage). Now, considering you're the new expert at FullCalendar, would you like to file a bug about making the view remembered by localStorage?

lcamacho:
Sure, np

PR #189, opened 2015-01-31
Bug 1117105 - Better ajax feedback

anurag90x:


airmozilla/manage/static/manage/js/survey-edit.js
```diff
@@ -31,19 +31,28 @@ $(function() {
     // hijack the ability to save/edit
     $('form.edit').submit(function() {
         var form = $(this);
+        $(form.children('button')).text('Saving..');
```
peterbe: Can you make that an ellipsis instead. `…`
It's a [Unicode character](https://en.wikipedia.org/wiki/Ellipsis) and it might not be clear how to copy that. But if I paste in what I can put in my editor it looks like this:
```javascript
$(form.children('button')).text('Saving…');
```

peterbe:
It's almost there. What happens though is that the save icon disappears. 
Before:
screenshot 2015-02-03 14 08 57
During:
screenshot 2015-02-03 14 09 42
After:
screenshot 2015-02-03 14 09 59

Where you do `form.children('button').text(...)` I would change that to `form.children('button span').text(...)` instead. 
And you'd need to change [these lines](https://github.com/mozilla/airmozilla/blob/master/airmozilla/manage/templates/manage/survey_edit_questions.html#L53-L56) to instead be something like:
```html
<button class="btn btn-primary btn-block">
    <i class="glyphicon glyphicon-floppy-disk"></i>
    <span>Save</span>
</button>
```

airmozilla/manage/static/manage/js/survey-edit.js
peterbe: Related but unrelated. You're here doing a `$(already_a_jquery_object)`. The code `form.children('button')` already is a "jQuery array" thing.

airmozilla/manage/static/manage/js/survey-edit.js
```diff
@@ -31,19 +31,28 @@ $(function() {
...
         var tr = form.closest('tr');
         var textarea = $('textarea', tr);
+        $(textarea).attr('readonly','readonly');
```
peterbe: Same here. `textarea` is already a jquery object.

airmozilla/manage/static/manage/js/survey-edit.js
```diff
@@ -31,19 +31,28 @@ $(function() {
...
         .then(function(response) {
+
+            $(textarea).removeAttr('readonly');
             if (response.error) {
                 alert('Unable to save. Error: ' + response.error);
             } else {
                 textarea.val(response.question);
+                $(form.children('button')).text('Saved');
             }
+            setTimeout(function()
```
peterbe: This might sound silly and picky but the style format we use everywhere else is:
```javascript
setTimeou(function() {
    ...
}, X * 1000);
```
I.e. after the `function()` there's always one space followed by a `{`.

peterbe:
@anurag90x Do you still have time to work on this? If not, I can finish it up.

anurag90x:
@peterbe  I've been pretty busy lately but I'll look at this tomorrow. Sorry about that!

PR #190, opened 2015-01-31
Fixes bug 1127938 - Update FullCalendar to latest version

lcamacho:


peterbe:
So, we need to change the back end. When I run this I get a 400 Bad Request on the URL:
`http://airmozilla.dev/calendar/data.json?start=2015-02-02T00%3A00%3A00-08%3A00&end=2015-03-14T00%3A00%3A00-07%3A00&_=1422921207335`

I think it's because before it used to send UTC timestamps. E.g. from prod
`https://air.mozilla.org/calendar/data.json?start=1422864000&end=1426316400&_=1422921225056`

So I think you need to change [CalendarDataForm](https://github.com/mozilla/airmozilla/blob/master/airmozilla/main/forms.py#L9-L35) to instead of taking in to timestamps, it should take in two datetimes.
Something like this might actually work:

```python
# UNTESTED
class CalendarDataForm(BaseForm):
    start = forms.DateTimeField()
    end = forms.DateTimeField()
    # same clean() as before
```
There should be tests also that you need to change.

peterbe:
Also, do you think you can look into making the buttons appear better?
screenshot 2015-02-02 15 59 43
I think they look strangely spaced together. It looks much better [on prod](https://air.mozilla.org/calendar/) where the buttons work as "pill boxes" (I *think* it's called). 

Also, when I hover over the buttons there's a strange gradient of blue. Can we remove that? Not only is it odd to have gradients like that, there's no other button that has that anywhere on the site, it's also new blue we don't use anywhere else I don't think.

I'm sorry if you're no expert at FullCalendar's CSS but I know you're clever so maybe you can find something. 
We can't edit the `fullcalendar.min.css` and we shouldn't but that's why we have the calendar.css file. Perhaps that file needs a complete refresher since it has hacks that were made based on the old FullCalendar.

airmozilla/main/forms.py
```diff
@@ -8,22 +6,18 @@
...
             raise forms.ValidationError(x)
 
     def clean_end(self):
         try:
-            return datetime.datetime.fromtimestamp(
-                self.cleaned_data['end']
-            )
+            return self.cleaned_data['end']
         except ValueError as x:
             raise forms.ValidationError(x)
```
peterbe: These two methods can now be deleted.

airmozilla/main/tests/test_views.py
```diff
@@ -1969,8 +1968,8 @@ def test_calendar_data(self):
         while last.month == first.month:
             last += datetime.timedelta(days=1)
 
-        first_ts = int(time.mktime(first.timetuple()))
-        last_ts = int(time.mktime(last.timetuple()))
+        first_ts = first.strftime('%Y-%m-%d')
+        last_ts = last.strftime('%Y-%m-%d')
```
peterbe: The "_ts" suffix came from the days when it was a "time stamp". We can just drop this now.

peterbe:
What has changed in this last commit?

By the way, I noticed lots of files that are now 755 instead of 644.

lcamacho:
List of changes:
- Fix event font color in Day and Week views
- Removed a couple of clean functions from form.py
- '_ts' suffix changed to something more appropiated in test_view.py
- Fix event time, now it shows the event in local time
- Fix permissions of FullCalendar assets

peterbe:
Just one thing now. 
screenshot 2015-02-03 13 59 45
We lost the formatting of the times. At least for me and my view. 
I see "9:30 MoFo Staff Call November". What's lacking is the am/pm notation. 
Now, I can guess that because it's not "**09:30** MoFo Staff Call November" I can guess it means that it's not a 24h clock. However, it's not clear.

lcamacho:
I'm using a 24 hours clock format like on production.

peterbe:
> I'm using a 24 hours clock format like on production.

You're right! It's bad on prod too. Can we do something about that?

lcamacho:
I can modify to show am/pm on month view and 12 hours clock without am/pm on week and day view, are you ok with this formats or it should show am/pm in every view?

peterbe:
I think it would be best with am/pm everywhere. 
Ideally, in some far future, we can use that detect.js library and make it "1:30 pm" for Americans and "13:30" for others.

peterbe:
r+ 
Well done! 
That's much easier to read. I'll merge it once travis goes green.

PR #192, opened 2015-02-05
Fix bug 1130144 - Save calendar view in localStorage

lcamacho:


lcamacho:
@peterbe because you have been so nice, here you have another pr

airmozilla/main/static/main/js/calendar.js
```diff
@@ -2,6 +2,10 @@ $(function() {
     var container = $('#calendar');
     var startsOnMonday = $('#startsOnMonday');
     var storage = localStorage.getItem('startsOnMonday');
+    var view = localStorage.getItem('view');
+    var save_view = function() {
```
peterbe: camelCase is the norm here. Eg. `startsOnMonday`, `initCalendar` etc. 
So then it should be `saveView`.

airmozilla/main/static/main/js/calendar.js
```diff
@@ -37,7 +42,10 @@ $(function() {
     startsOnMonday.on('change', function(){
         localStorage.setItem('startsOnMonday', startsOnMonday.prop('checked')? 1 : 0);
         container.fullCalendar('destroy');
+        view = localStorage.getItem('view');
         initCalendar();
+        $('.fc-right button').on('click', save_view);
```
peterbe: This doesn't belong here.

airmozilla/main/static/main/js/calendar.js
```diff
@@ -37,7 +42,10 @@ $(function() {
...
     });
     initCalendar();
+    $('.fc-right button').on('click', save_view);
```
peterbe: I think there's a better way to do it. Hang on...

peterbe: http://fullcalendar.io/docs/display/viewRender/
"Triggered when a new date-range is rendered, or when the view type switches."

lcamacho: I did a test with that method but the problem with that is there's no way to figure what cause the render event, so for example I can't differentiate between the first render and a render caused by an user who clicked on one of the buttons to change the view.

airmozilla/main/static/main/js/calendar.js
```diff
@@ -37,7 +42,10 @@ $(function() {
     startsOnMonday.on('change', function(){
         localStorage.setItem('startsOnMonday', startsOnMonday.prop('checked')? 1 : 0);
         container.fullCalendar('destroy');
+        view = localStorage.getItem('view');
         initCalendar();
+        $('.fc-right button').on('click', saveView);
```
peterbe: You're creating an onclick handler every time.

airmozilla/main/static/main/js/calendar.js
```diff
@@ -37,7 +42,10 @@ $(function() {
...
     });
     initCalendar();
+    $('.fc-right button').on('click', saveView);
```
peterbe: So I saw your other comment and it's a fair point but I've look into this in detail and I think we should use [eventAfterAllRender](http://fullcalendar.io/docs/event_rendering/eventAfterAllRender/).
It gets fired every time you load the calendar (after the events in it have been rendered). 
It *also* gets fired when you switch the view. 

(it's almost the same as `viewRender` but happens slightly later)

It sucks that there's no way of distinguishing between initial rendering and manual-view-switching, but what sucks even more is depending on the hack of knowing the exact CSS class of the area where the buttons are. 

Your trick is clever and has advantages but it also means it's a hack and it might stop working at some future revision. It's generally much better to stick to the documented public API. 

So, would you mind changing to that. That means it's writing to the localStorage every time you open the calendar. But I guess so be it.

airmozilla/main/static/main/js/calendar.js
```diff
@@ -31,12 +33,16 @@ $(function() {
...
+            },
+            eventAfterAllRender: function(view) {
+                localStorage.setItem('view', view.name);
             }
         });
     };
     startsOnMonday.on('change', function(){
         localStorage.setItem('startsOnMonday', startsOnMonday.prop('checked')? 1 : 0);
         container.fullCalendar('destroy');
+        view = localStorage.getItem('view');
```
peterbe: What is this for? If it's for keeping the same view when re-recreating the calendar then we'd use `view = container.fullCalendar('getView')` no?

lcamacho: that's right I will update the pr

PR #193, opened 2015-02-06
Fix bug 1130504 - Fix calendar buttons on Chrome turn blue when focused

lcamacho:


peterbe:
Tested. Works like a charm!

PR #194, opened 2015-02-06
fixes bug 1129371 - Set author in feed(s) [as used on Planet Mozilla]

peterbe:


airmozilla/main/views.py
```diff
@@ -963,6 +975,9 @@ def item_link(self, event):
                 return self._get_webm_link(event)
         return self._root_url + reverse('main:event', args=(event.slug,))
 
+    def item_author_name(self, event):
+        return self.title
```
lcamacho: Shouldn't this be the event's creator?

peterbe: There's no such thing really. Our "creator" field is more administrative.

PR #195, opened 2015-02-09
Bug 1089713 - Unfinished push for review and comments

bugzPDX:
I will change the name of test_starred.py to test_views.py to match the library naming convention once it's closer to being complete.


airmozilla/starred/templates/starred/home.html
```diff
@@ -0,0 +1,48 @@
...
+      {% if prev_page_url %}
+        <li class="next">
+          <a href="{{ prev_page_url }}{% if tags %}?{{ tags_query_string(tags) }}{% endif %}">
+            {{ _('Newer videos') }}
+          </a>
+        </li>
+      {% endif %}
+    </ul>
+  </nav>
+  {% endif %}
```
peterbe: All of this is now repeated code from `main/templates/main/home.html` and we mustn't forget to turn all of this into a macro.

airmozilla/starred/tests/test_starred.py
```diff
@@ -0,0 +1,198 @@
+import json
```
peterbe: Name of this file should be `test_views.py`. Some apps, like `manage` has so many views that `test_views.py` became so large it needed to be exploded into smaller files. That's why things like `manage/tests/test_views.py` became things like `manage/tests/views/test_something.py`.

airmozilla/starred/tests/test_starred.py
```diff
@@ -0,0 +1,198 @@
...
+        structure = json.loads(response.content)
+        eq_(structure, {"ids": []})
+
+        # add an event id to the list
+        structure['ids'].append(events[0].id)
+        # send synced list to browser
+        response = self.client.post(
+            url,
+            content_type='application/json',
+            data=json.dumps(structure)
```
peterbe: We probably won't send it as `application/json`. Instead we'll send it as `multipart/form-data` and the test code will be short and sweat:
```python
response = self.client.post(url, ids=structure['ids'])
```
and in the view where this is being POSTed to you simply do:
```
ids = request.POST.getlist('ids')
```

airmozilla/starred/tests/test_starred.py
bugzPDX: Yup I noted that I'd change it once I was closer to being done with it. The temporary name just keeps me less confused as I move between files.

airmozilla/starred/tests/test_starred.py
bugzPDX: That is very handy!

airmozilla/starred/tests/test_starred.py
```diff
@@ -0,0 +1,198 @@
...
+        )
+        # get the list and verify it was updated
+        structure = json.loads(response.content)
+        eq_(structure, {"ids": [events[0].id]})
+
+    def test_remove_starred_event(self):
+
```
peterbe: This test is called "test_remove_something_something" so then it should ideally be about testing only the remove part. 
One way of doing that is to shortcut to the necessary condition. E.g.:
```python
def test_removed_starred_event(self):
    event1_id = self.events[1].id
    StarredEvent.objects.create(user=self.user, event=self.events[1])
    StarredEvent.objects.create(user=self.user, event=self.events[2])
    response = self.client.post(url, {'ids': [self.events[1].id]})
    assert something on the response
    ok_(not StarredEvent.objects.filter(id=event1_id))
    
```

Do you see? This way, it's only testing one thing; the delete view.

airmozilla/starred/tests/test_starred.py
```diff
@@ -0,0 +1,198 @@
...
+            data=json.dumps(structure)
+        )
+        # get the list and verify it was updated
+        structure = json.loads(response.content)
+        eq_(structure, {"ids": []})
+
+
+    def test_anonymous_user(self):
+        # log out user from setUp()
+        self.client.logout()
```
peterbe: uh?

airmozilla/starred/tests/test_starred.py
```diff
@@ -0,0 +1,198 @@
...
+from funfactory.urlresolvers import reverse
+from nose.tools import eq_, ok_
+
+from airmozilla.main.models import Event
+from airmozilla.starred.models import (
+    StarredEvents,
+)
+
+
+class TestStarredEvent(TestCase):
```
peterbe: Change the base class from `django.test.TestCase` to `airmozilla.base.tests.testbase.DjangoTestCase`.

airmozilla/starred/tests/test_starred.py
```diff
@@ -0,0 +1,198 @@
...
+
+
+    def setUp(self):
+        super(TestStarredEvent, self).setUp()
+
+        # create the url
+        self.url = reverse('starred:sync_starred_events')
+        # so, let's sign in
+        User.objects.create_user('lisa', password='secret')
+        assert self.client.login(username='lisa', password='secret')
```
peterbe: Once you've changed the test case base class you'll be able to use `self._login()` to get a user to be created and logged in to the `self.client`.

airmozilla/starred/tests/test_starred.py
```diff
@@ -0,0 +1,198 @@
...
+            title='The Best Event',
+            slug='event3',
+            description=event.description,
+            start_time=event.start_time,
+            privacy=Event.PRIVACY_PUBLIC,
+            placeholder_img=event.placeholder_img,
+            location=event.location,
+        )
+        # get events from the database
+        self.events = Event.objects.all()
```
peterbe: This whole setUp is getting too big and creating too much stuff. Fixtures are tricky but basically they should be kept at a very small minimum. The `'Test event'` sample event that is loaded from `airmozilla/manage/tests/main_testdata.json` is used so incredibly much that it was worth it so it's an exception. 

Basically, if there's something in here that isn't needed for every (or 95%) test in the class, then it's excess. 

The reasoning is twofold:

1. For someone reading the test, it's better to be explicit (try typing `import this` in any python interpreter shell) about what data is available so as to remove any guesswork. If you can see the `Event.objects.create(...)` about 5 lines away from where you're reading code it's easier to comprehend what's going on.

2. One day we might remove one of the tests but in the rush miss to clean up the necessary fixture that was used by only that one last test. At that point it's really confusing why the fixture is needed. And taking the time to do clean-up refactoring is rare. This is the same predicament when you add stuff to a .css file needed by some widget and much later in time you delete the widget but fail to clean up the .css file and then that file just grows and grows. 

Does that make sense?

airmozilla/starred/urls.py
```diff
@@ -0,0 +1,9 @@
+from django.conf.urls import patterns, url
+
+from . import views
+
+urlpatterns = patterns(
+    '',
+    url(r'^$', views.starred_events, name='starred_events'),
+    url(r'^sync$', views.sync_starred_events, name='sync_starred_events'),
```
peterbe: We should almost always put a trailing slash. E.g. `r'^sync/$'`

airmozilla/starred/views.py
```diff
@@ -0,0 +1,54 @@
...
+from airmozilla.main.models import Event
+from .models import StarredEvents
+from airmozilla.base.mozillians import fetch_user_name
+
+
+@json_view
+def sync_starred_events(request):
+    if request.method == 'POST':
+        decode = json.loads(request.body)
+        ids = decode['ids']
```
peterbe: Yeah, let's change this to 
```python
ids = request.POST.getlist('ids')
```

airmozilla/starred/views.py
```diff
@@ -0,0 +1,54 @@
...
+                    user=request.user,
+                    event=event
+                )
+            except Event.DoesNotExist:
+                # ignore events that don't exist but fail on other errors
+                pass
+
+
+    starred = StarredEvents.objects.filter(user=request.user)
+    ids = [star.event.id for star in starred] #Better way to do this?
```
peterbe: Yeah. You can use something like this:
```
ids = starred.values_list('event_id', flat=True)
```
I think.

airmozilla/starred/tests/test_starred.py
bugzPDX: I was just starting to write this test when I pushed my branch for you to look at.

airmozilla/starred/tests/test_starred.py
bugzPDX: Yup, that makes sense

airmozilla/starred/tests/test_starred.py
peterbe: ignore my comment :)

bugzPDX:
@peterbe take a peek at this whenever you get a chance.

bugzPDX:
More for review @peterbe

peterbe:
Great progress. One thing that is tricky is the whole thing about getting curated groups to appear there correctly. I need to think of a better way so that no code has to be duplicated in `starred/views.py` and `main/views.py`. You've got plenty of things to work on so I'm not going to stress you on how to create template macros. ...yet. So I'll help with that this week.

bugzPDX:
When you say template macros do you mean pulling the code for listing views
and for pagination out and making separate templates for those? I can work
on that for sure but should those two sections be 'includes' or 'extends'?
I was thinking includes but I don't know enough about templating. I see
that lots of templates extend from main_base.html so pulling that code out
will be interesting. Also, how the heck do I fake a placeholder_image in my
event creation??

bugzPDX:
For you to look over @peterbe

airmozilla/main/templates/main/home.html
```diff
@@ -11,7 +11,50 @@
...
+    <a class="star star-off"></a>
+    <p>
+      {{ short_desc(event) | safe_html }}
+      <a class="go" href="{{ href }}">{{ _('See more') }}</a>
+    </p>
+  </div>
+</article>
+{% endmacro %}
 
+{% macro event_paginate(next_page_url, tags, prev_page_url) %}
```
peterbe: Make `tags` the last argument so that `next_page_url` and `prev_page_url` can be together. 
Also, since `tags` is something `None` or `''` you might want to write it as `tags=None` in the parameter list. 
I think for the starred events won't ever have tags.

peterbe:
I think we need to name these macros something more descriptive. When you see `article(event)` in the template and don't spot the macro definition you might really struggle to understand where this callable is coming from. 

Is it a context variable? Is it from the context processors? Is it a helper function?
Those are the 99.9% most common places where things come from. 

What do you think if we rename `article` to `article_macro` instead? I think that would really help a developer who is debugging this and does a grep on the word `article_macro`.

airmozilla/main/templates/main/main_base.html
```diff
@@ -148,6 +149,7 @@
   <script src="{{ static('js/libs/include.js') }}"></script>
   <script src="{{ static('browserid/api.js') }}"></script>
   <script src="{{ static('browserid/browserid.js') }}"></script>
+  <script src="{{ static('starred/js/star_event.js') }}"></script>
```
peterbe: It might sound unimportant but if you put this *above* `js/libs/include.js` this code will execute slightly ahead of the other stuff related to persona. 
We want that stuff to take effect really quickly so that the user (a user with "quick eyes") doesn't notice that her starred events start out as a hollow star and then suddenly changes to a filled star.

airmozilla/starred/models.py
```diff
@@ -0,0 +1,12 @@
...
+from django.db import models
+from django.contrib.auth.models import User
+from airmozilla.main.models import Event
+
+
+class StarredEvent(models.Model):
+    event = models.ForeignKey(Event)
+    user = models.ForeignKey(User)
+    created = models.DateTimeField(auto_now_add=True)
+
```
peterbe: We should put up an additional guard that prevents there from ever existing duplicates here that only differ in their `created` value. Check out the django documentation on how to use `unique_together` in a `Meta` class on the model.

airmozilla/starred/static/starred/js/star_event.js
```diff
@@ -0,0 +1,5 @@
+$(document).ready(function(){
```
peterbe: Use the `$(function() {` notation instead. You'll find it being used in many of the other jquery-using files in the project.

airmozilla/starred/static/starred/js/star_event.js
```diff
@@ -0,0 +1,5 @@
...
+    $(".entry-summary").on("click", function () {
```
peterbe: I don't think this is true. If you click anywhere on the `<div class="entry-summary">` it shouldn't toggle the star.
I think what you're looking for is `$('.entry-summary').on('click', 'a.star', function() {`

airmozilla/starred/static/starred/css/star_event.css
```diff
@@ -0,0 +1,18 @@
+.star {
```
peterbe: Let's be explicit that this is always an `a` tag. So use `a.star` instead.

airmozilla/starred/static/starred/css/star_event.css
```diff
@@ -0,0 +1,18 @@
...
+    background: url('../img/Star24YellowTransparent.png') no-repeat;
+    height: 24px;
+    width: 24px;
+    background-position: 0 0;
+    position: absolute;
+    left: 140px;
+    top: 70px;
+    opacity: 0.9;
```
peterbe: At the moment we have, in the HTML, this: `<a name="star star-off"></a>`  which, due to the lack of a `href` attribute, does get the pointer cursor when you over the mouse over it. So add to this CSS:
```css
cursor: pointer;
```

airmozilla/starred/templates/starred/home.html
```diff
@@ -0,0 +1,16 @@
...
+{% set page = page|default('home') -%}
+
+{% block content_main %}
+  {% for event in events %}
+  {{ article(event) }}
+  {% endfor %}
+  <!-- OR
+  {% for star in starred %}
+  {{ article(star.event) }}
+  {% endfor %}
```
peterbe: Yes! This is much better. That way we don't need to make a list of events in the view function.

airmozilla/starred/views.py
```diff
@@ -0,0 +1,91 @@
...
+            'starred:home',
+            args=(starred_paged.next_page_number(),)
+        )
+    if starred_paged.has_previous():
+        prev_page_url = reverse(
+            'starred:home',
+            args=(starred_paged.previous_page_number(),)
+        )
+
+    events = [s.event for s in starred_paged]
```
peterbe: No longer needed.

airmozilla/starred/views.py
```diff
@@ -0,0 +1,91 @@
...
+        .order_by('name')
+    )
+    for event_id, name in curated_groups:
+        curated_groups_map[event_id].append(name)
+
+    def get_curated_groups(event):
+        return curated_groups_map.get(event.id)
+
+    context= {
+        'events': events,
```
peterbe: no longer needed

airmozilla/starred/templates/starred/home.html
bugzPDX: Great! This was my preference.

airmozilla/main/templates/main/home.html
bugzPDX: @peterbe if I move tags it breaks pagination.

peterbe: You'll also need to change the two places where this macro function is called.

airmozilla/main/templates/main/_side_events.html
```diff
@@ -19,6 +19,7 @@ <h4 class="entry-title">
             </span>
             {{ event.title }}
           </a>
+        <a class="star sidebar" data-id="{{ event.id }}" title="Save to watch later"></a>
```
peterbe: Let's call it "Save as a starred event" or something. "watch later" is a different terminology.

airmozilla/main/templates/main/home.html
```diff
@@ -33,6 +32,7 @@ <h3 class="entry-title">
       <div class="entry-summary">
         <a href="{{ href }}">
           {{ show_thumbnail(live, '160x90') }}
+        <a class="star live" data-id="{{ event.id }}" title="Save to watch later"></a>
```
peterbe: ditto

airmozilla/starred/static/starred/js/star_event.js
```diff
@@ -0,0 +1,96 @@
...
+    console.log(oJson);
+
+
+   $('#content').on('click', 'a.star', function () {
+      Stars.toggleArrayPresence($(this).data('id'));
+      $(this).toggleClass('star-on');
+   });
+
+   Stars.sync();
+});
```
peterbe: This file has a bunch of 3-space indentation I noticed. It should be 4 spaces.

airmozilla/main/templates/main/home.html
```diff
@@ -1,5 +1,5 @@
 {% extends 'main/main_base.html' %}
-{% set front_page = (events.number == 1) %}
+{% set front_page = (events and events.number == 1) %}
```
peterbe: This is breaking `airmozilla/main/tests/test_views.py:TestPages.test_channels_page` because it now requires that there are events on this page for it to be considered the "front page". 
A channel page, might not have any events (it might [have sub-channels](https://air.mozilla.org/channels/mozfest/)) but it might still be the front page. 

Why did we need this again? The purpose of this "front_page" business is two-fold. To list the live events if there are any and to show the channel title/banner.

airmozilla/starred/templates/starred/home.html
```diff
@@ -0,0 +1,36 @@
...
+    $(function() {
+    if (Stars.getStars().length) {
+            $('#events').load(window.location + '',
+                $.param({ids: Stars.getStars()}, true),
+                function () {
+                    Stars.triggerClickedStars();
+                }
+            );
+        } else {
+            $('#events').text("Sorry. You have no starred events");
```
peterbe: We need to fix this. First of all, "Sorry" isn't really right. It's nothing to be sorry about. :)

Secondly, I think we should avoid any English inside the Javascript (except some rare crazy error message that shouldn't happen e.g (`console.error("Seems your localStorage is broken. File a bug please");`)

What you should do is add a DOM element to the page. I'm thinking something like this:
```html
<div id="events">
    <p class="has-stars">Please wait. Loading your starred events...</p>
    <p class="no-stars" style="display:none">You have no starred events. Yet.</p>
</div>
```
Then, instead of using `$('#events').text(...)` we do something like 
```javascript
if (length) {
    $('#events').load(...);
} else {
    $('#events .has-stars').hide();
    $('#events .no-stars').show();  
}
```

peterbe:
Here's a problem. If I have multiple `<a class="star" data-id="123">` on the same page, you can get "out of sync". 
This can easily happen on the Starred page if one of the events I've starred is in the current sidebar. 
screenshot 2015-03-17 15 25 39

It's also not unrealistic that on the home page, the most recent event is in the sidebar because it's Trending. 
In fact, it's happening right now on air.mozilla.org
screenshot 2015-03-17 15 27 48
(note the "Brown Bag - EU SURVEILLE Project: Ethics, Privacy and Anonymity" event appears twice)

I think the solution is simple. Untested but something like this could work
```javascript
// instead of
Stars.toggleArrayPresence($(this).data('id'));
$(this).toggleClass('star-on');
// do
var id = $(this).data('id');
Stars.toggleArrayPresence(id);
$('a.star[data-id="' + id + '"]').toggleClass('star-on');
```

airmozilla/starred/views.py
```diff
@@ -0,0 +1,102 @@
...
+    if ids:
+        # how do we order_by the order in ids?
+        events = Event.objects.filter(id__in=ids).order_by('-created')
+        template_name = 'starred/events.html'
+
+    elif request.user.is_authenticated():
+        events = (
+            Event.objects.filter(starredevent__user=request.user.id)
+            .order_by('-starredevent__created')
+        )
```
peterbe: The sort order is wrong. It should have the most recently starred appear first. I.e. `'starredevent__created'`. I think.

peterbe:
One thing that wonder about is when I unstar an event when I'm on the Starred page. (Signed in or not). 
What [Youtube does](https://www.youtube.com/playlist?list=WL) is that when you click the "Remove" icon to remove one of your Watch Later, it removes it from the list. Should we perhaps do the same? Should we not worry about it for now?

airmozilla/main/templates/main/_paginate.html
```diff
@@ -0,0 +1,20 @@
...
+{% if prev_page_url %}
+<li class="next">
+<a href="{{ prev_page_url }}{% if tags %}?{{ tags_query_string(tags) }}{% endif %}">
+{{ _('Newer videos') }}
+</a>
+</li>
+{% endif %}
+</ul>
+</nav>
+{% endif %}
```
peterbe: Can you indent this HTML?

airmozilla/starred/static/starred/js/star_event.js
```diff
@@ -0,0 +1,99 @@
...
+                        stars = stars.slice(0, 500);
+                        localStorage.setItem('stars', JSON.stringify(stars));
+                        Stars.triggerClickedStars();
+                    }
+                });
+            }
+        },
+
+        triggerClickedStars: function () {
+                $('a.star').each(function(i, element) {
```
peterbe: indentation for this whole function.

peterbe:
The file `test_views.py` is very freely mixing `"` and `'`. E.g. 
```python
eq_(structure, {'csrf_token': csrf_token, "ids": []})
```
I suggest you stick to `'` everywhere. The only place I use `"` is when I write English things that a human might one day see :)

peterbe:
Ok. I think I can see the bigger picture now with regards to `star_event.js`. 
What I suggest you do is revamp it so that your `Stars` module only exposes three functions:
* `initialSync`
* `toggleArrayPresence` (this needs a better name too!)
* `getStars`

You can make the current `sync` function a private function. I.e. inside the closure. The only actors that needs to ever call `sync` is `initialSync` and `toggleArrayPresence`.

peterbe:
I.e. I mean it'd look like this instead:
```javascript
Stars=(function() {
  function sync(...)
  return {
      initialSync: function() {
        localStorarage? sync();
      }, 
      toggleArrayPresence: function() {
          localStorage! sync();
      }
  }
})()
```
and in `main_base.html`:
```html
<script>
$(function() {
    Stars.initialSync();
});
</script>
```

peterbe:
By the way, with regards to the last comment. There are a couple more functions the `Stars` module would expose by the way. All for the sake of the `starredevents.js`  code.

peterbe:
Also, I like the idea of the `setPostSyncCallback` (except I think a better name would be `registerPostSync` instead). I think that's going to work. 
I still don't understand the `if ($('#nav-main a.current').attr('href') === '/starred/') {` statement but in theory it all makes sense. 

One thing I don't like about `EventLoader` is that it repeats the code `!!$('starred-event').length`. I think you'd need to add another function to the `Stars` module so that `EventLoader` can do `if (Stars.isSignedIn()) {` instead.

airmozilla/main/templates/main/_events.html
```diff
@@ -0,0 +1,24 @@
...
+    </h2>
+    {% set curated_groups = get_curated_groups(event) %}
+    {% include 'main/_event_privacy.html' %}
+  </header>
+  <div class="entry-summary">
+    <p class="event-date">{{ event.start_time|js_date }}</p>
+    <a href="{{ href }}">
+      {{ show_thumbnail(event) }}
+    </a>
+    <a class="star" data-id="{{ event.id }}" title="Save as a starred event"></a>
```
peterbe: This needs to be:
```
<a class="star{% if star_on %} star-on{% endif %}" ...
```

airmozilla/starred/views.py
```diff
@@ -0,0 +1,105 @@
...
+    def get_curated_groups(event):
+        if events:
+            return curated_groups_map.get(event.id)
+
+    context = {
+        'events': starred_paged,
+        'get_curated_groups': get_curated_groups,
+        'next_page_url': next_page_url,
+        'prev_page_url': prev_page_url,
+    }
```
peterbe: Add
```
'star_on': True
```

PR #198, opened 2015-02-10
Fix bug 1130519 - Add selenium test for calendar page

lcamacho:


dev-requirements.txt
```diff
@@ -3,3 +3,4 @@ pep8==1.6.1
 pyflakes==0.8.1
 mccabe==0.3
 flake8==2.3.0
+selenium==2.44.0
```
peterbe: Move this to `requirements.txt` because that's where we have all other dependencies for running tests. 
Like `mock` and `nose` etc. 

This file (`dev-requirements.txt`) are for extra things that you don't have to necessarily run to make a patch. However, it's just highly highly recommended that you check your code with these tools. But not a must. 

What do you think, should we split up `requirements.txt` into two files? One for running the server and one for running tests? E.g. `requirements.txt` (to be used on servers) and `test-requirements.txt` to be used by travis and local developers.

I like the separation because it means less things needs to be installed on the production servers but it's also nice that there's just one big file you need to look at.

peterbe: These are the packages that you only need for testing but not for running the server. Not a very big list. 
```
nose==1.3.4
mock==1.0.1
pyquery==1.2.8
```

lcamacho: I think it's better to have just one file plus with selenium they are just 4 packages

lcamacho:
I just have to force a new travis build because of WebDriverException("Can not connect to GhostDriver") and sometimes fails at local

peterbe:
Can I ask a humble favor. I like that you rebase your PRs before I merge them. That eliminates debugging/hacking commits and stuff. 
However, I think it's better to do the rebase at the very last moment when the code is almost ready to be merged. 
As I'm reviewing, when the patch is large it's often useful to see what has changed between revisions of the PR's lifecycle. 

So feel free to pepper the PR with more and more commits as you go and then when you get a "r+" you can rebase it into one commit.

peterbe:
I'm really excited about this! Once we have this up and running we start adding more tests and perhaps develop some neat shortcuts and patterns. 

I bulldozered in with some ideas of my own with: https://github.com/lcamacho/airmozilla/pull/1
What do you think?

peterbe:
Hmm... I restarted the Travis test. It timed out. I wonder if that's the strange errors you had seen too. 
If this is happening so often in travis then I wonder if we dare to have these tests on by default in travis.

lcamacho:
For what I've seen so far most of my fails are because phantomjs is not seeing anything after clicking on calendar link but that happens on linux where is at version 1.9 but on mac where latest version is 2.0 it happens less frecuently.

peterbe:
Strange that Coveralls says the coverage goes down 0.03% when we add more tests :)

peterbe:
This is great. But we need one last thing and that's a mention in the README about how to run these tests. The README should mention setting `RUN_SELENIUM_TESTS = True` and that you need to installed `phantomjs`. I don't think we need to explain how to install phantomjs because it's different on different operating systems.

README.md
```diff
@@ -365,6 +365,9 @@ Tests and test coverage
 Included is a set of comprehensive tests, which you can run by:
 ``./manage.py test``
 
+If you want to run selenium tests don't forget to add ``RUN_SELENIUM_TESTS = True``
+in your ``airmozilla/settings/local.py`` file also you need to install [phantomjs](http://phantomjs.org/).
+
```
peterbe: It's Markdown. Not reStructuredText. So you can just use a single back-tick. 
Also, try to keep the sentences shorter than 80 characters like the rest of the file. It's not fatal if it spills over 80 chars but that should only really be for certain code snippets that are hard to split onto multiple lines. 

Also, it's called `PhantomJS` not `phantomjs`.

peterbe:
So just one little nit there, but: r+

Would you mind rebasing it into one commit now (with the nit incorporated) and I'll merge it.

PR #209, opened 2015-02-20
Improved editorconfig

lcamacho:


.editorconfig
```diff
@@ -1,15 +1,9 @@
...
-indent_size = 4
-
-[*.js]
-indent_style = space
-indent_size = 4
-
-[*.css]
-indent_style = space
-indent_size = 4
+root = true
```
peterbe: Why this?
What does this do?

lcamacho: From editorconfig page
root ; ... Set to "true" to stop .editorconfig files search on current file.

peterbe: So basically, it doesn't search in my `~` directory because all configs are here now thanks to the `[*]`. 
Awesomeness!

PR #219, opened 2015-03-05
Additional dependencies for running airmozilla

RGaonkar:
In order to tackle the errors while installing packages mentioned in requirements.txt using the pip command, I have added the necessary package names and the installation procedure.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ dependencies. You do this with:
 
     pip install -r requirements.txt
     pip install -r dev-requirements.txt
-
+These two files are available in the airmozilla respository, after you clone it from github
 The second file is necessary so you can
 
 **Step 2 - Get the code**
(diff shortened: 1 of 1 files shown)


peterbe:
Thanks but I don't think this is particularly useful. 

First of all, I think it's very common to put the requirements.txt file inside the repository and it wouldn't ever be placed outside it. 

Also, the instructions you added for using `sudo apt-get install ...` are only applicable if you're using a form of Linux that has `apt-get`. If you could re-organize it so that it's under a "troubleshoot" section or something specifically for people on Linux with `apt-get` available. 

Again, I'm sorry if it's hard to install. This repo is no different than many other python based projects. I'm not sure the solution is to add deep explanations of python basics but instead try to find some good links for people who are new to installing python dependencies.

peterbe:
I think it's valuable to explain to Linux users what packages are expected to be installed. Can you re-organize it so that what you have added is instead under its own section called "Installation on Linux"?

PR #220, opened 2015-03-05
About page added.

shivamMg:
- About template added to main templates
- Updated **urls.py**, **main_base.html** and **views.py** accordingly


peterbe:
I'm afraid this is not the right solution. 

The reason the About page is a static page that can be managed in the management pages is because that way the content can be edited in the brower without having to edit the source code. 

I think this is not going to be solved easily. Perhaps once you've solve any other [good first bug](https://bugzilla.mozilla.org/buglist.cgi?resolution=---&status_whiteboard_type=allwordssubstr&query_format=advanced&status_whiteboard=[good%20first%20bug]&bug_status=NEW&component=Air%20Mozilla&product=Webtools&list_id=12065632) you'll be more up-to-speed and I can help you explain why this bug is actually quite hard to solve.

shivamMg:
I think you are right. Probably should have started with the easy ones. This one looks easy. https://bugzilla.mozilla.org/show_bug.cgi?id=1124274

peterbe:
Yeah, that's a great start.

peterbe:
The most important thing, for you, right now, is to get warmed up with the whole development cycle of creating git branches, making PRs, addressing PR feedback etc. And running tests.

PR #221, opened 2015-03-06
Update event.html

aarshvig:
1154/1158 passed. @peterbe Please have a look.


peterbe:
Hi! Good start. However, as you can see you'll need to do some work on the way you moved that HTML. If you're careful and check exactly how the "if statements" are laid out you'll see you probably pasted something in to the wrong place or something. 
E.g.
```python
# From this
if 'one'==1:
    print "header"
video()
if 'two'==2:
    print "other stuff"
```

```python
# To this
video()
if 'one'==1:
    print "header"
    if 'two'==2:
        print "other stuff"
```
when the correct solution would be this:
```python
video()
if 'one'==1:
    print "header"
if 'two'==2:
    print "other stuff"
```

peterbe:
Also, ping me on IRC so I can show you how you can run just the tests that are failing so you don't have to wait 50+ seconds for running the the whole test suite every time.

peterbe:
We did it in [this commit](https://github.com/mozilla/airmozilla/commit/34fbdf415d49717c66c7d538be9a701027f96677) instead

PR #224, opened 2015-03-07
Update fields.py

aarshvig:
@peterbe please have a look


airmozilla/main/fields.py
```diff
@@ -1,7 +1,7 @@
 from django.db import models
 from django.forms.fields import Field
 from django.forms.util import ValidationError
-from django.utils import simplejson as json
+import json
```
peterbe: @aarshvig because this is now your second patch, it's time to start getting more and more perfect. 

So, what you need to do is to move this line up to the top of the file. There's a stylistic pattern. You'll find this in almost every other file. It goes like this:
```
import some
import standard
import python
import library
[EMPTY LINE]
import some
import third_party
import python
import library
[EMPTY LINE]
import things
import from
import django
[EMPTY LINE]
import third
import party
import django
import related
import libraries
[EMPTY LINE]
import airmozilla
import specific
import modules
import and
import libraries
```

Look at [main/views.py](https://github.com/mozilla/airmozilla/blob/master/airmozilla/main/views.py#L1-L56) for example.

Does that make sense?

peterbe:
Also, before this can be merged the git commit message needs to be improved. On your last pull request, I manually corrected this for you but now you're more experienced so I can ask more from you :)

What you need to do is write a short git commit message that looks something like this:
```
fixes bug 1140641 - django.utils.simplejson is deprecated
```
The word "fixes " at the beginning isn't always necessary but very often it is. Then it's "bug XXXXX" which is the bugzilla ID. Then it's " - " followed by a short description which either mentions what the bug was or what it corrects. 

If you look at the [recent history of commits](https://github.com/mozilla/airmozilla/commits/master) you'll see a pattern.

peterbe:
Hi!
I think what you don't know is that you can continue to work on the same PR without having to open a [new one](https://github.com/mozilla/airmozilla/pull/228).

GitHub is amazing like that. Suppose you want to change the commit message, you can do that by rebasing. Rebasing is an advanced subject but once you've done it once it gets really quite easy. 

To rebase, you'll use the interactive mode so you can re-write the commit message for example. 

I use this all the time when I start on a branch to experiment with something and then later decide I want to do it properly with a bug ID and everything. Then I use `git rebase -i master` to start the editor so I can re-write the commit message. 

When you rebase you often have to do a "forced push" back into your PR. 

The possibly best way to achieve this is to try and if you get stuck have a backup so you can start over. Also, doing it at a time when there are people on IRC who can help you is a good idea.

PR #232, opened 2015-03-10
Fix bug 1141269 - Thousands commas on Dashboard numbers

blossomica:
I believe I fixed this bug.
You can have a look and let me know if anything needs to be changed. 
Thanks.


peterbe:
Works wonderfully!

blossomica:
Great!

PR #233, opened 2015-03-10
Fix bug 1141298 - Deprecate UserFindForm

blossomica:
I've fixed this bug. 
If anything is wrong you can let me know.

### DIFF
--- a/airmozilla/manage/views/users.py
+++ b/airmozilla/manage/views/users.py
@@ -39,9 +39,7 @@ def users(request):
             User.objects.filter(_mozilla_email_filter).count()
         ),
     }
-    form = forms.UserFindForm()
     context = {
-        'form': form,
         'users_stats': users_stats,
     }
     return render(request, 'manage/users.html', context)
--- a/airmozilla/manage/forms.py
+++ b/airmozilla/manage/forms.py
@@ -78,17 +78,6 @@ class GroupEditForm(BaseModelForm):
         model = Group
 
 
-class UserFindForm(BaseForm):
-
-    email = forms.CharField(max_length=200)
-
-    def clean_email(self):
-        email = self.cleaned_data['email']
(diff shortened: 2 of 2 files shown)


airmozilla/manage/views/users.py
```diff
@@ -39,7 +39,7 @@ def users(request):
             User.objects.filter(_mozilla_email_filter).count()
         ),
     }
-    form = forms.UserFindForm()
+    form = forms.EventEditForm()
```
peterbe: You might want to review that. It's not right.

blossomica:
All right, can you explain bit more on what's the problem with it?

blossomica:
Hi @peterbe , I tried to make some changes, I might be wrong again I am not sure. This is my first time working with django so I guess I need to read more on it to improve because I find it hard in some way. Any suggestions of things I can read will be appreciated.

airmozilla/manage/views/users.py
```diff
@@ -39,12 +39,7 @@ def users(request):
             User.objects.filter(_mozilla_email_filter).count()
         ),
     }
-    form = forms.UserFindForm()
-    context = {
-        'form': form,
-        'users_stats': users_stats,
-    }
-    return render(request, 'manage/users.html', context)
```
peterbe: Keep the `context` variable. But just remove the `'form': form,` line. (and the `form = forms.UserFindForm()` line of course)

blossomica:
Oh okay, that's pretty much what I did in my last pull request [https://github.com/blossomica/airmozilla/commit/21b7743b2fdae4cdd2da5e63154acac8f914db49] but I didn't keep the context variable. So I will add the variable back.

peterbe:
Wondeful! You're in!

blossomica:
Awesome!

PR #239, opened 2015-03-14
fixes Bug 1141286 - List video meta data as a neat table

aarshvig:
@peterbe please have a look.


peterbe:
Thanks. I'm looking forward to this but I think it's going to require a bit more work. 

To start with, why is there no changes to `event.html`?

The event page is precious and I'm going to raise the bar really high on this page. It's the most common and the most important page. As important as the home page. Reason is that most traffic comes to this page by a big margin. People who land on an event page expects that page to be really really fast and we need to do everything we can to make sure the video starts to load as soon as possible. 

I don't like the use of bootstrap. Bootstrap is an amazing CSS framework but it's heavy. Also, we already have a CSS framework in place on the main site. It'd be different if bootstrap is all we had but to add 27Kb CSS just for one little tabbed view is unfortunately not acceptable. I bet that if you research it deeper you'll find that it's possible to build a tabbed view without much css. A quick google found [this](https://css-tricks.com/examples/CSSTabs/radio.php) which looks like a good start. 

With regards to you making a contribution for the sake of the Outreachy application. Don't stress about completion of the PR. Effort, passion and attention to quality counts more than being in the commit history or not.

aarshvig:
http://picpaste.com/pics/Screenshot_from_2015-03-24_08_29_04-OQBhpE6M.1427166159.png
@peterbe 
I have implemented this using pure css. Do you require any changes?

aarshvig:
@peterbe http://picpaste.com/pics/Screenshot_from_2015-03-26_00_13_35-CLv35bUe.1427326486.png
?

peterbe:
That's it! That looks good. Looking forward to see how you solve this with CSS and javascript.

PR #244, opened 2015-03-19
fixes bug 1132523 - Open Graph fields for non-event pages

blossomica:
Hi,

for starters I tried to change the open graph fields for home, my thought was that it should link to home and display airmozilla logo, however something here is not  right, the first bug it has is the link gives me an error saying - 'views' is undefined - so that means the way I have linked to home is wrong. I am also not sure if the rest are okay. So some feedback and guidance on this will be extremely helpful.

Also I am thinking of removing  the content from `<meta name="keywords" content="{{ ','.join(tags) }}">` 
and adding the content "home" instead because home doesn't really have tags to join I suspect. 

Thanks you.


airmozilla/main/templates/main/home.html
```diff
@@ -11,7 +11,17 @@
 </style>
 {% endblock %}
 
-
+{% block extra_head %}
+{{ super() }}
+<meta name="keywords" content="{{ ','.join(tags) }}">
```
peterbe: This is not an Open Graph meta tag.

airmozilla/main/templates/main/home.html
```diff
@@ -11,7 +11,17 @@
...
+<meta property="og:title" content="Home">
```
peterbe: This should be the same title as this page which isn't Home. It's "Air Mozilla | Mozilla, in Video" but you shouldn't have to write that twice. There's got to be a way to write this one and use in multiple places.

airmozilla/main/templates/main/home.html
```diff
@@ -11,7 +11,17 @@
...
+<meta property="og:url" content="{{ url('main:home', views.home) }}">
```
peterbe: This needs to be made into an absolute URL.

airmozilla/main/templates/main/home.html
```diff
@@ -11,7 +11,17 @@
...
+<meta property="og:description" content="Airmozilla Homepage') }}">
```
peterbe: Copy the text from https://air.mozilla.org/about/
Also, as you can see this isn't valid template code.

airmozilla/main/templates/main/home.html
```diff
@@ -11,7 +11,17 @@
...
+{% set thumb = "{{ static('img/logo-airmo-md.png') }}" %}
```
peterbe: `static()` returns a URL. Not a thumbnail object.

airmozilla/main/templates/main/home.html
blossomica: Yes you are right, It does not have an og property so it is a normal meta-tag, I'll remove that.

airmozilla/main/templates/main/home.html
blossomica: Yes, there is a leftover parenthesis, I'll use that description and I'll fix that.

blossomica:
I'll work with the feedback you have given me and do some updates.  

Thanks so much for the detailed feedback.

airmozilla/main/templates/main/home.html
blossomica: Wow... I totally confused that part. I was using the tab title instead. I now got it, thanks.

blossomica:
@peterbe - I did some changes. Whenever you get the chance you can let me know if this is okay or there is something wrong. If this is okay then I will proceed on making the other pages. 

Is there anyway to test this and see how it looks like when shared?

airmozilla/main/templates/main/home.html
```diff
@@ -11,7 +11,16 @@
 </style>
 {% endblock %}
 
-
+{% block extra_head %}
+{{ super() }}
+<meta property="og:title" content="{{ page_title }}">
+<meta property="og:url" content="{{request.build_absolute_uri(None)}}">
+<meta property="og:description" content="Air Mozilla is the Internet multimedia presence of Mozilla, with live and pre-recorded shows, interviews, news snippets, tutorial videos, and features about the Mozilla community." }}">
```
peterbe: `}}` at the end.

airmozilla/main/templates/main/home.html
```diff
@@ -11,7 +11,16 @@
 </style>
 {% endblock %}
 
-
+{% block extra_head %}
+{{ super() }}
+<meta property="og:title" content="{{ page_title }}">
+<meta property="og:url" content="{{request.build_absolute_uri(None)}}">
```
peterbe: No need for the `None` part. And use the same spacing after the `{{` and `}}`.

peterbe:
I see you've rebased and squashed this into one commit. Cool. But it's too soon to do that. 

Ideally, what you should do is just keep piling on commits into the pull request until you get the final "go ahead" or "r+" from me. When you get that, I'll usually remind you and say something like "Great! It works perfectly. Can you squash this down to 1 commit now please?"

That way, I can see what the incremental changes are during the review process.

blossomica:
Oh okay. I didn't know that I have to rebase at the end alone. I will be doing that from now on. Thanks for letting me know. I am glad I have been able to get to this point. I didn't expect being able to work on this and I hardly even knew what open graph tags are, but I am glad that I do now.  
I'll correct my mistakes and start working on another page.

blossomica:
1. I did some changes you can check on whenever you get the chance, and even though I get no errors when running airmozilla locally, the content of `"og:image:width"  "og:image:height"  "og:title"` comes off empty when I do view page source which makes me wonder if the code I wrote to get those contents is actually right. 

2. Also the other non-event pages seem to use the same code so am I suppose to make a new file with that portion of code and be importing it instead of repeating it? Is that possible? 

[not closely related to this bug]
3. I noticed different indentation patterns on templates like the home.html, event.html and the calendar.html pages. I see the README says airmozilla is structured using PEP 8 and that we should read about that + pyflakes, but seeing these pages being structured kind of different makes me wonder what is the right code pattern to follow. Can you please enlighten me on that? Do I follow the pattern I see on each page separately or use a specific style guide no matter how the page is currently structured?

airmozilla/main/templates/main/home.html
```diff
@@ -11,7 +11,16 @@
 </style>
 {% endblock %}
 
-
+{% block extra_head %}
+{{ super() }}
+<meta property="og:title" content="{{ page_title }}">
+<meta property="og:url" content="{{ request.build_absolute_uri() }}">
+<meta property="og:description" content="Air Mozilla is the Internet multimedia presence of Mozilla, with live and pre-recorded shows, interviews, news snippets, tutorial videos, and features about the Mozilla community.">
+{% set thumb = thumbnail('img/logo-airmo-md.png', '200x200', crop='center') %}
```
peterbe: This doesn't work. It's not easy to know but you can't generate thumbnails from static images. Only for images that are uploaded as part of the data. 
Instead I would just opt to finding out the width and height is of that logo and typing that in. E.g. `<meta property="og:image:width" content="222">`

And for the `og:image` you need to use a function that returns a static URL that is absolute. I think a good example of this is here: https://github.com/mozilla/airmozilla/blob/master/airmozilla/roku/templates/roku/categories.xml
`abs_static`.

blossomica:
@peterbe - Thanks, I fixed those. I used view-source to see if the function 'abs_static' and the rest worked properly and it seems okay. When I went to check the dimensions of the logo I notice the "airmozilla" wording is in white, there is one logo with black wording but it is in larger size so I think I will have to ask Richard to make another medium sized one with black wording to make sure it appears when shared? The large one is 730 x 160 pixels and I know most social medias have white background behind the shared item. Apart from this `+<meta property="og:title" content="{{ page_title }}">` returns blank content so I guess I am doing something wrong there even though I get no actual error when I run airmozilla.

peterbe:
That looks great. However, I think we should ask @RAMilewski to create a dedicated logo at is at least 200px (or bigger) wide and 200px (or bigger) and with solid color background.

RAMilewski:
Is this just the AirMo logo, or the AirMo Logo with the wordmark?     What's the preferred background color?

blossomica:
@RAMilewski - The one I was using is one with AirMo Logo with the wordmark, the wordmark was white. The whole logo was in a transparent background so I thought if someone shares the page at a place like facebook the wordmark might not appear at all if the background the logo appears on is white. I think @peterbe can tell us what is the preferred background and the rest.

RAMilewski:
Talked with Peter and it sounds like having an oversized image is better than having one that's too small.
Try this and see how it works.    Let me know if you want changes.

airmozilla_logo

blossomica:
@RAMilewski - Thanks for the logo. I've downloaded it and will try it tomorrow (almost 2am my side). I'll let you know if there is a need any changes. I appreciate the effort.

blossomica:
@peterbe - I did some changes so when you get the chance you can check them and let me know if it is okay. I tried to get the page title without repeating it in the code, but I am not sure if my fix is correct so If it's not i'd definitely appreciate having some feedback on it. Thank you.

peterbe:
Excellent! I'll take a proper look later today. 
It looks like it's going to be fine. Would you mind now to rebase this down to just 1 commit?

blossomica:
@peterbe - Thanks :) I just rebased it to just 1 commit, once you have a look let me know if it is okay. We have other non event pages ie. About, Channels, Calendar and Tag Cloud, that have no open graph tags as well, I haven't fix those yet.

peterbe:
> About, Channels, Calendar and Tag Cloud, that have no open graph tags as well, I haven't fix those yet. 

Good point. Perhaps we should move your changes from `home.html` to `main_base.html` so it's used on every page instead. Every page other than the event pages. Yeah, let's do that. Would you mind?

blossomica:
Okay I will try to figure out how to do that.

blossomica:
@peterbe - I think I just fixed that. I have not rebase this new change, if it is okay and I get the go ahead from you I will do so. You can let me know if I have to fix anything else or if I have done anything wrong.

airmozilla/main/templates/main/main_base.html
```diff
@@ -16,8 +16,15 @@
 Air Mozilla | Mozilla, in Video
 {% endblock %}
 
+
 {% block extra_head %}
 <link rel="alternate" title="{{ feed_title }}" href="{{ feed_url }}" type="application/rss+xml">
+<meta property="og:title" content="{{ self.page_title() }}">
+<meta property="og:url" content="{{ request.build_absolute_uri() }}">
+<meta property="og:description" content="Air Mozilla is the Internet multimedia presence of Mozilla, with live and pre-recorded shows, interviews, news snippets, tutorial videos, and features about the Mozilla community.">
+<meta property="og:image" content="{{ abs_static('img/logo-airmo-md.png') }}">
```
peterbe: `og` not `md`

peterbe:
Sooo close. We just need to reference the right image and then this'll be ready.

blossomica:
I just fixed it, you can have a look.

peterbe:
Hmm.. This is what it would look like on Facebook. https://www.dropbox.com/s/ysrfbv01ckxemn9/Screenshot%202015-03-24%2016.25.55.png?dl=0

I'm going to see what the output is if I change to a less wide image.

peterbe:
screenshot 2015-03-24 16 29 14

I tried chopping it but it now looks worse :)
@RAMilewski can you chop us a new image? One that is 470px wide (or slightly bigger) and 246px high (or slightly higher)

blossomica:
After reading this  "Use images that are at least 1200 x 630 pixels for the best display on high resolution devices. "  and "If your image is smaller than 600 x 315 px, it will still display in the link page post, but the size will be much smaller." (source: https://developers.facebook.com/docs/sharing/best-practices#images ) I think the height might have been the problem but I am not sure.

peterbe:
Indeed. 1200x630 is about 2:1 whereas the [one Richard prepared](https://github.com/blossomica/airmozilla/blob/bug1132523-open-graph-fields-for-non-event-pages/airmozilla/base/static/img/logo-airmo-og.png) is about 4:1.

RAMilewski:
It's hard to tell exactly what's going on there. Here's one that's 1200 px wide and it comes out 351 px high.   I could do on that's 1200 x 630, but that would have a lot more black for the same amount of logo.

The last one was 800 x 235.  It could be that the height being under their 315 minimum triggered something strange.  Let's see how this looks.   If it fails too, we can try a different layout with the 1.91:1 aspect ratio.
airmozilla_logo_1200x351

RAMilewski:
Here's one that is exactly 1200 x 630, just in case.    ...although if we can make it work I would prefer the horizontal layout above.
airmozilla_logo_1200_630

blossomica:
@RAMilewski - Thanks for the new logo.
@peterbe - Do you want me to push this logo here or you can check on it without me pushing it then if it is okay I can add this logo and delete the previous one?

peterbe:
@blossomica replace the one in the PR with the latest one Richard put there. 
It's perfect @RAMilewski

RAMilewski:
airmozilla_logo_1200_630
This is a slightly modified version of the last one.  The advantage to this one is that if it gets cropped as a square, then it still works.

blossomica:
@peterbe - I just did the change and I used the newer better version @RAMilewski did. I am not sure why Travis says there are merge conflicts. If there is any problem you can let me know, if it is okay then you can also let me know if it is time for me to rebase it. 

@RAMilewski - Thanks so much for the help. It looks great :)

peterbe:
The reason it says it can't be merged is because your work on the home.html template has since changed since you started. Basically, a more recent change has landed in master that conflicts with this PR. What you're going to need to do is a rebase. Do you know how to do that? If not, I can guide you on IRC.

PR #247, opened 2015-03-22

swetarani:
@peterbe  I have tried to present the metadata of events in a simple tabular form. Please see to it and comment if any improvements required.


peterbe:
Note that @aarshvig already got started on something similar in https://github.com/mozilla/airmozilla/pull/239
However, her solution uses a tabbed view which might look nice because you don't have to show everything at first and thus you can afford to put a bit more information in without worrying about making the page too noisy.

PR #252, opened 2015-03-26
Fixes bug 1147898 - Stringify URL and event IDs

bugzPDX:


airmozilla/starred/static/starred/js/starredevents.js
```diff
@@ -7,7 +7,7 @@ var EventLoader = (function() {
                 container.html($( '.no-stars' ).clone().toggleClass('no-stars').show());
             }
             else {
-                var url = window.location + '';
+                var url = window.location.toString();
```
peterbe: Actually, I think this slipped in an earlier review comment. You shouldn't use `location`. You should use `location.pathname`. Its value would be something like `/starred/`. 
If you visit `http://localhost:8000/starred/?foo=bar#anchor` then you'll be setting `url` to `http://localhost:8000/starred/?foo=bar#anchor` which is excessive. What you only want is `/starred/`. 

And you then don't need any `.toString()` trick since `location.pathname` is always a string and not an object.

airmozilla/starred/static/starred/js/star_event.js
```diff
@@ -60,7 +60,7 @@ var Stars = (function() {
             if (signedIn) {
                 // then the <starred-event> tag would have the list of ids
                 var serverStars = [];
-                var ids = $('starred-event').data('ids');
+                var ids = $('starred-event').data('ids').toString();
```
peterbe: yay!

airmozilla/starred/static/starred/js/starredevents.js
bugzPDX: Oh good to know!

PR #255, opened 2015-03-27
Fixes bug 1147902 - No initial EventLoader.load() for signed in

bugzPDX:


peterbe:
Would you mind rebasing this. I introduced a bug on the master branch that breaks some test suites.

peterbe:
Actually, it matter not.

PR #259, opened 2015-03-27

aarshvig:
@peterbe please check this pr.


airmozilla/main/templates/main/event.html
```diff
@@ -139,75 +139,108 @@ <h2 class="event-date">
         </p>
       {% endif %}
       <p>{{ event.description | carefulnl2br | safe_html }}</p>
+      <div class="tabs">
 
+      {% if event.additional_links %}
+      <div class="tab">
+        <input type="radio" id="addlinks" name="tab-group-1">
```
peterbe: All of these `id="something"` can you change them to something like `id="tab-something"`.
Some names are too generic that it could interfere with other things. For example `id="tags"` is very generic.

peterbe:
It looks really good! I'm impressed!

There's some polishing to do...

* For me, the first time I loaded it, it made the "Tags" tab (the second one) the highlighted tab. I think the first one ("Additional Links") is probably the most important one. 

* The CSS has different indentation for different selectors. I see some are indented with 2 spaces, some indented with 6 spaces but the general pattern for that file was 4 spaces ([which is correct](https://github.com/mozilla/airmozilla/blob/master/.editorconfig#L5)).

* The spacing between the tabs is too short. If you don't look careful some tabs look like one big tab. E.g. "Views Since Archived Embed"

* There is a little imperfection on the left-hand side by the first tab 
screenshot 2015-03-27 14 50 51

* I think the embed button stopped working. Look at [a page like this](https://air.mozilla.org/vr-cinema-meetup-3/) where you can see that "Small (640x380)" and "Big (896x504)" are next to each other.

airmozilla/main/static/main/css/event.css
```diff
@@ -95,3 +95,62 @@ p.pending {
...
 }
+.tabs{
+    position: relative;
+    min-height: 200px;
+    clear: both;
+    margin: 25px 0;
+}
+.tabs {
+  position: relative;
+  min-height: 200px; /* This part sucks */
```
peterbe: why does it suck?

airmozilla/main/static/main/css/event.css
```diff
@@ -95,3 +95,62 @@ p.pending {
...
+  clear: both;
+  margin: 25px 0;
+}
+.tab {
+      float: left;
+}
+.tab label {
+      background: #20272a;
```
peterbe: Isn't that supposed to be `background-color`?

airmozilla/main/static/main/css/event.css
```diff
@@ -95,3 +95,62 @@ p.pending {
...
+      left: 1px;
+}
+.tab [type=radio] {
+      display: none;
+}
+.content {
+  position: absolute;
+  top: 28px;
+  left: 0;
+  background: #2d373b;
```
peterbe: same here

aarshvig:
@peterbe Changed id to "tab-number"

airmozilla/main/templates/main/event.html
```diff
@@ -139,75 +139,108 @@ <h2 class="event-date">
...
-        <div class="embed hidden">
-          <a href="#" class="current">Small (640x380)</a>
-          <a href="#">Big (896x504)</a>
-          <br>
-          <textarea rows="1" cols="80" class="small-embed-code">{{ embed_code }}</textarea>
-          <textarea rows="1" cols="80" class="big-embed-code hidden">{{ embed_code_big }}</textarea>
+        <div class="tab">
+          <input type="radio" id="tab-6" name="tab-group-1">
+          <label for="tab-6">Embed</label>
+          <div class="content">
```
peterbe: I think that if you just add `embed` here it'll work. So
```html
<div class="content embed">
```

airmozilla/main/static/main/css/event.css
```diff
@@ -95,3 +95,62 @@ p.pending {
...
+.content {
+  position: absolute;
+  top: 28px;
+  left: 0;
+  background-color: #2d373b;
+  right: 0;
+  bottom: 0;
+  padding: 20px;
+  border-bottom-right-radius: 5px;
+  border-bottom-left-radius: 5px;
```
peterbe: Lets remove the bottom rounded corner. I think it would look strange if the tab thing has rounded corners and the sidebar thing below doesn't.

airmozilla/main/static/main/css/event.css
```diff
@@ -95,3 +95,62 @@ p.pending {
...
+.tab {
+      float: left;
+}
+.tab label {
+      background: #20272a;
+      padding: 8px;
+      border: none;
+      margin-left: 0px;
+      position: relative;
+      left: 1px;
```
peterbe: This is what causes that little imperfection on the left-hand side.

peterbe:
Excellent progress. Now, the only things I can find that needs some attention are:

1. The color of the inactive tabs is a strange new color. It's too bright to indicate that is the **non-active** tab. Try to find a share of grey that is between the active tab background color and the whole-page background color. Does that suggestion make sense?

2. The spacing under the writing on the tabs is a bit too small. I think it would make sense to use the same amount of padding **over** the text as it is underneath. I.e. put the text in a vertical center. 

3. Can you make clicking the tabs change the cursor. I think you can easily do this by adding `cursor: pointer;` to the `.tab label` selector. 

4. The CSS is style poorly style :) Weird to say that but check your indentation. It's supposed to be 4 spaces for each key-value pair.

peterbe:
So what's left here? A better background color for in-active tabs? And the fixup of the CSS structure.

airmozilla/main/static/main/css/event.css
```diff
@@ -95,3 +95,62 @@ p.pending {
...
+    border-top-right-radius:5px;
+    position: relative;
+    cursor:pointer;
+}
+.tab [type=radio] {
+    display: none;
+}
+.content {
+    position: absolute;
+    top: 28px;
```
peterbe: Change this to 32px instead.

peterbe:
We need to fix this:
screenshot 2015-04-01 16 02 43

peterbe:
And this (where there are multiple lines of additional links):

screenshot 2015-04-01 16 03 20

peterbe:
Lastly regarding the background color of the in-active tab, I suggest you simply remove the background color entirely. It'll then look like this which I think is good enough
screenshot 2015-04-01 16 04 35

peterbe:
It actually doesn't grow vertically but that's OK I guess. It looks great. r+

Would you mind rebasing it into just one nicely formatted and named git commit?

PR #261, opened 2015-03-28
Fixes bug 1148566 - Concat serverStars to stars

bugzPDX:
@peterbe serverStars were being ignored.

### DIFF
--- a/airmozilla/starred/static/starred/js/star_event.js
+++ b/airmozilla/starred/static/starred/js/star_event.js
@@ -70,6 +70,8 @@ var Stars = (function() {
                 // jquery array comparison http://stackoverflow.com/a/7726509/205832
                 if (!($(stars).not(serverStars).length === 0 &&
                       $(serverStars).not(stars).length === 0)) {
+                    stars = stars.concat(serverStars);
+                    triggerClickedStars();
                     sync();
                 }
             }
--- a/airmozilla/starred/templates/starred/events.html
+++ b/airmozilla/starred/templates/starred/events.html
@@ -1,2 +1,6 @@
-{% include 'main/_events.html' %}
-{% include 'main/_paginate.html' %}
+{% if events %}
+    {% include 'main/_events.html' %}
+    {% include 'main/_paginate.html' %}
+{% else %}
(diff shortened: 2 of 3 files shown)


airmozilla/starred/static/starred/js/star_event.js
```diff
@@ -70,6 +70,8 @@ var Stars = (function() {
                 // jquery array comparison http://stackoverflow.com/a/7726509/205832
                 if (!($(stars).not(serverStars).length === 0 &&
                       $(serverStars).not(stars).length === 0)) {
+                    stars = stars.concat(serverStars);
```
peterbe: This makes me nervous. What if `stars` is not empty or if there is an item in both arrays. Then you'd get repeated IDs in there. 
Why not do something like this (untested):
```javascript
serverStars.forEach(function(id) {
    if (stars.indexOf(id) === -1) { 
        stars.push(id); /// Should that be unshift() instead??!?
    }
});
```

airmozilla/starred/templates/starred/events.html
```diff
@@ -1,2 +1,6 @@
-{% include 'main/_events.html' %}
-{% include 'main/_paginate.html' %}
+{% if events %}
+    {% include 'main/_events.html' %}
+    {% include 'main/_paginate.html' %}
+{% else %}
+    You have no starred events. Yet.
```
peterbe: Any text should always be in a tag. E.g. `<p>You have...`

airmozilla/starred/static/starred/js/star_event.js
bugzPDX: Yes they could get repeated but the server is going to sort it all out. I think changing it is just going to slow everything down for not a lot of benefit.

peterbe: A ha! I get it. The `stars` gets set to `stars = response.ids;` a couple of milliseconds later.

peterbe:
@bugzPDX Oops! I merged this too quickly. Would you mind addressing the comment I made about the "You have no starred events. Yet" text being in a `<p>` tag?

PR #266, opened 2015-04-01
Fixes bug 1147657 - Remove position: absolute from a.star.event

bugzPDX:


airmozilla/main/templates/main/event.html
```diff
@@ -98,8 +98,8 @@
      <header class="entry-header">
       <h1 class="entry-title">
         {{ event.title }}
-      </h1>
       <a class="star event" data-id="{{ event.id }}" title="Save as a starred event"></a>
```
peterbe: This is `.entry-title a.star` which is here: https://github.com/bugzPDX/airmozilla/blob/1147657/airmozilla/starred/static/starred/css/star_event.css#L31

airmozilla/starred/static/starred/css/star_event.css
```diff
@@ -9,7 +9,6 @@ a.star {
     cursor: pointer;
 }
 a.star.event {
-    position: absolute;
     left: 130px;
     top: 64px;
```
peterbe: Shouldn't these be deleted too?

PR #267, opened 2015-04-02
Fixes bug 1150690 - Show loading message only when loading starred events

bugzPDX:


airmozilla/starred/templates/starred/home.html
```diff
@@ -12,10 +12,13 @@ <h2 class="section-title">Starred Events</h2>
 {% if events %}
     {% include 'starred/events.html' %}
 {% else %}
-    <p class="has-stars">Please wait. Loading your starred events...</p>
+    <p >You have no starred events. Yet.</p>
 {% endif %}
 </div>
-    <p class="no-stars" style="display:none">You have no starred events. Yet.</p>
+<div style="display:none">
+    <p class="no-stars">You have no starred events. Yet.</p>
```
peterbe: Now we have repeated the same string in two places. That violates the DRY principle. 

I think we need a deeper solution here. 
Basically, the fault, I think, lies in how the AJAX get replaces all the contents of the `<div id="events">` tag. When it does that it destroys English text which we might end up needing later.

peterbe:
An alternative implementation https://github.com/mozilla/airmozilla/pull/268

PR #268, opened 2015-04-03
fixes bug 1150690 - "Please wait" on Starred page when signed in

peterbe:
@bugzPDX r?


peterbe:
So the key ingredient here is...:

1) The "Please wait" is only ever shown once if you arrive at the page with nothing. 

2) When the post-sync-callback is called and there are no events, the container is hidden.

bugzPDX:
@peterbe r+ on this. Thank you!

PR #269, opened 2015-04-03

aarshvig:
@peterbe Check this out :)

### DIFF
--- a/airmozilla/main/templates/main/event.html
+++ b/airmozilla/main/templates/main/event.html
@@ -98,8 +98,8 @@
      <header class="entry-header">
       <h1 class="entry-title">
         {{ event.title }}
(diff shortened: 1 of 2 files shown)


airmozilla/main/templates/main/event.html
```diff
@@ -98,8 +98,8 @@
      <header class="entry-header">
       <h1 class="entry-title">
         {{ event.title }}
-      <a class="star event" data-id="{{ event.id }}" title="Save as a starred event"></a>
       </h1>
+      <a class="star event" data-id="{{ event.id }}" title="Save as a starred event"></a>
```
peterbe: This `<a>` tag is supposed to be **inside** the `<h1>` tag.

PR #270, opened 2015-04-06
Fixes bug 1147582 - Add test for starred event pagination

bugzPDX:


airmozilla/starred/tests/test_views.py
```diff
@@ -227,3 +227,53 @@ def test_anonymous_starred_events(self):
...
+
+        for i in range(25):
+            event = self.create_event("Event Title %d" % (i + 1))
+            StarredEvent.objects.create(user=user, event=event)
+            events.append(event.id)
+
+        response = self.client.get(url)
+        eq_(response.status_code, 200)
+
+        for i in range(len(events)):
```
peterbe: change this to `for id in events:`

airmozilla/starred/tests/test_views.py
```diff
@@ -227,3 +227,53 @@ def test_anonymous_starred_events(self):
...
+            match = 'data-id="%d"' % events[i]
+            if i < 10:
+                self.assertTrue(match in response.content, i)
```
peterbe: Instead of `self.assertTrue` we use the "shorthand" tool which is `ok_()`.
And the anti-equivalent of `ok_(...)` is `ok_(not ...)`.

airmozilla/starred/tests/test_views.py
```diff
@@ -227,3 +227,53 @@ def test_anonymous_starred_events(self):
...
+            else:
+                self.assertFalse(match in response.content, i)
+
+        ok_('href="/starred/page/2/"' in response.content)
```
peterbe: Can you make that use the `reverse()` function instead. E.g. 
```python
ok_('href="%s"' % reverse('starred:home', args=(2,)) in response.content)
```

airmozilla/starred/tests/test_views.py
```diff
@@ -227,3 +227,53 @@ def test_anonymous_starred_events(self):
...
+        ok_('href="/starred/page/0/"' not in response.content)
+
+        response = self.client.get(url + 'page/2/')
```
peterbe: Again, here use `reverse()`.

airmozilla/starred/tests/test_views.py
```diff
@@ -227,3 +227,53 @@ def test_anonymous_starred_events(self):
...
+
+        for i in range(len(events)):
+            match = 'data-id="%d"' % events[i]
+            if i >= 20:
+                self.assertTrue(match in response.content, i)
+            else:
+                self.assertFalse(match in response.content, i)
+
+        ok_('href="/starred/page/2/"' in response.content)
+        ok_('href="/starred/page/4/"' not in response.content)
```
peterbe: How is this third section different from the second one?

peterbe:
Some nits to work on there but great work!

peterbe:
@bugzPDX Any chance you can have a look at the above mentioned nits? If not, I can take over this PR.

bugzPDX:
Yes I am in progress on this one.

airmozilla/starred/tests/test_views.py
bugzPDX: The second section checks to make sure there is a next page and a previous page. The third section is different in that it checks to make sure there is only a previous page and not a next page.

bugzPDX:
@peterbe would you give me direction on the lines that are too long? I messed with them for a while but everything I tried looked awful.

airmozilla/starred/tests/test_views.py
```diff
@@ -227,3 +227,53 @@ def test_anonymous_starred_events(self):
...
+
+        for id in range(len(events)):
+            match = 'data-id="%d"' % events[id]
+            if id < 10:
+                ok_(match in response.content, id)
+            else:
+                ok_(match not in response.content, id)
+
+        ok_('href="%s"' % reverse('starred:home', args=(2,)) in response.content)
+        ok_('href="%s"' % reverse('starred:home', args=(0,)) not in response.content)
```
peterbe: To avoid making these lines too long, you can perhaps just set, before these two lines, something like:
```python
output = response.content
```
That's 10 characters less. Then the lines don't get too long.

peterbe: Yes, the strict pep8 rules are silly but it's nice that there are rules at all. It makes it "un-debateable" about personal preferences on style and it makes the code quite clean and sustainable when it's set to a high quality.

airmozilla/starred/tests/test_views.py
```diff
@@ -227,3 +227,53 @@ def test_anonymous_starred_events(self):
...
+
+        for i in range(25):
+            event = self.create_event("Event Title %d" % (i + 1))
+            StarredEvent.objects.create(user=user, event=event)
+            events.append(event.id)
+
+        response = self.client.get(url)
+        eq_(response.status_code, 200)
+
+        for id in range(len(events)):
```
peterbe: The `for thing in range(len(sequence)):` is known as an anti-pattern in Python. 
Almost always you should just iterate on the object itself. E.g. `for thing in sequence:`

However, if you want to the index numbers when iterating over a list you use `enumerate()`. E.g.

```python
list1 = ['a','b','c']
list2 = ['d', 'e', 'f']
for i, item in list1:
    print (item, list2[i])
```

peterbe: In fact, you ought to be able to do the above code with a list of events and not a list of ids. E.g.

```python
for i in range(25):
    events.append(event)
...
for event in events:
    match = 'data-id="%d"' % event.id
    ...
```

peterbe:
That's excellent! r+

Would you mind rebasing it into 1 commit and then I'll merge?

bugzPDX:
Ok it should be good to go.

PR #284, opened 2015-04-23
Bug 1157440 - More prominent display of start time for upcoming events.

blossomica:
I made a fix for this, I am not sure if It looks the way it was envisioned to look like so If there is a need for any changes here  you can let me know.  I forgot to name the PR "fixes bug ..." but If that's a problem I can fix that. I realized it after seeing the message on github.

### DIFF
--- a/airmozilla/main/templates/main/event.html
+++ b/airmozilla/main/templates/main/event.html
@@ -89,6 +89,23 @@
         {{ event.title }}
       <a class="star event" data-id="{{ event.id }}" title="Save as a starred event"></a>
       </h1>
+
+      {% if event.is_upcoming() %}
+      <h4>
+         <p>
+            {{ event.start_time | js_date }}
+            {% if event.location %}
+             &mdash;
+             {{ event.location_time.strftime("%I:%M%p") }}
+              {% if not event.location.name.startswith('Cyberspace') and not 
+                 event.location.name.startswith(settings.DEFAULT_PRERECORDED_LOCATION[0]) %}
+                 in {{ event.location.name }}
+              {% endif %}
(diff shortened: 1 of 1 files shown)


airmozilla/main/templates/main/event.html
```diff
@@ -89,6 +89,21 @@ <h1 class="entry-title">
...
+             &mdash;
+             {{ event.location_time.strftime("%I:%M%p") }}
+              {% if not event.location.name.startswith('Cyberspace') and not 
+                 event.location.name.startswith(settings.DEFAULT_PRERECORDED_LOCATION[0]) %}
+                 in {{ event.location.name }}
+              {% endif %}
+            {% endif %}
+           {% endif %}
+         </p>
+      </h4>
```
peterbe: If `{% if event.is_upcoming() %}` is false, doesn't that mean you end up with
```html
<h4>
  <p>
  </p>
</h4>
```

blossomica:
@peterbe - Yes you are right. I didn't think of that part but I fixed it, thanks. You can let me know if it is okay now or it needs any other changes.

peterbe:
r+

blossomica:
@peterbe - Do I always have to wait for Travis to finish scanning the last pull request after I get r+ when the difference is so slight that I could just rebase it and then let Travis run a scan on the rebased final file instead?

peterbe:
No, as soon as you get the r+ you can rebase and force push into the branch even if travis hasn't finished the un-squashed version, so to say.

blossomica:
@peterbe - All right, that's good to know. I rebased it.

PR #291, opened 2015-04-29
fixes bug 1157938 - Hide Call info on archived events

blossomica:
Umm, this is half complete. I decided to work on this till I figure out how to write tests. I have been having a hard time to find where is "call info" is in the event edit form. I see it calls bootstrapform(form), I went to the main/forms.py to see if it is in there but then again I can't really find its actual location to edit. It seems to be embedded at event_edit through some other file I haven't been able to locate yet; so that is one place I got stuck. Also I don't know if call info tab is to appear for pending events. Lastly what is the purpose of call info? It seems like a tag but maybe it does something special which I haven't figure that out? @peterbe

### DIFF
--- a/airmozilla/main/forms.py
+++ b/airmozilla/main/forms.py
@@ -61,6 +61,8 @@ class EventEditForm(BaseModelForm):
         self.fields['picture'].label = (
             'Select an existing picture from the gallery'
         )
+        if not (self.event.is_upcoming() or self.event.is_live()):
+            del self.fields['call_info']
 
     def clean(self):
         cleaned_data = super(EventEditForm, self).clean()
--- a/airmozilla/main/templates/main/event.html
+++ b/airmozilla/main/templates/main/event.html
@@ -188,7 +188,7 @@
         </div>
       {% endif %}
 
-      {% if event.call_info %}
+      {% if event.call_info and (event.is_upcoming() or event.is_live()) %}
         <div class="tab">
           <input type="radio" id="tab-4" name="tab-group-1">
           <label for="tab-4">Call Info</label>
--- a/airmozilla/main/tests/test_views.py
+++ b/airmozilla/main/tests/test_views.py
@@ -1178,6 +1178,34 @@ class TestPages(DjangoTestCase):
             in response.content
         )
 
+    def test_call_info_presence(self):
+        event = Event.objects.get(title='Test event')
+        event.call_info = 'More info'
+
+        url = reverse('main:event', args=(event.slug,))
+
+        event.archive_time = None
+        event.start_time = timezone.now() + datetime.timedelta(days=1)
+        event.save()
+        assert event.is_upcoming()
+        response = self.client.get(url)
+        eq_(response.status_code, 200)
+        ok_('More info' in response.content)
+
+        event.start_time = timezone.now()
+        event.save()
+        assert event.is_live()
+        response = self.client.get(url)
+        eq_(response.status_code, 200)
+        ok_('More info' in response.content)
+
+        event.archive_time = timezone.now()
+        event.save()
+        assert not event.is_live() and not event.is_upcoming()
+        response = self.client.get(url)
+        eq_(response.status_code, 200)
+        ok_('More info' not in response.content)
+
     @mock.patch('airmozilla.manage.vidly.urllib2.urlopen')
     def test_event_with_vidly_token_urlerror(self, p_urlopen):
         # based on https://bugzilla.mozilla.org/show_bug.cgi?id=811476


airmozilla/main/templates/main/event.html
```diff
@@ -188,7 +188,7 @@ <h1 class="entry-title">
         </div>
       {% endif %}
 
-      {% if event.call_info %}
+      {%  if event.call_info and event.is_live() or event.is_upcoming() %}
```
peterbe: Extra space has crept in there between the `%` and the `i`.

peterbe:
Many questions :)
Let's see...

* The purpose of this is if you're going to have a presentation over Vidyo (our video conferencing software) you here enter information about the virtual room you're going to be using. That tells the people who hook up the live stream with the Vidyo software what to do. 

* The presence of it, to view, is only applicable if it's upcoming **or** live. So not pending or archived.

* The form class to dig into is [EventEditForm](https://github.com/mozilla/airmozilla/blob/master/airmozilla/main/forms.py#L38-L63). You're going to need to dig into that `__init__` method and do an if statement on the `self.event` and possible do a `del self.fields['call_info']`

blossomica:
Thanks for the explanations; it made things clearer. I am working on it. I'll post a link to this PR on bugzilla once I fix the other part of the code that is missing.

blossomica:
@peterbe  This works for me so I don't understand where is the error it says from lint.sh . I see no error when I run it on my side.

peterbe:
So Air Mozilla has a strict pep8 and pyflakes checker on almost all code. It's a bit silly but it's something to stick to. 
What I have, is a checker that checks all the files I edit before I git commit. That prevents me from ever (well, you can override it) pushing badly formatted code into travis. 

Put [this file](https://gist.github.com/peterbe/bba51ea999a3a1e36cc5) into the `.git/hooks/` directory and call it `pre-commit`. Make sure it's executable so you have to run something like this:

```
cd myprojects/airmozilla
wget https://gist.githubusercontent.com/peterbe/bba51ea999a3a1e36cc5/raw/f93769f0bf5e6a57e03fc942eee3c185cd7c7e31/pre-commit
mv pre-commit .git/hooks/
chmod +x .git/hooks/pre-commit
```

Also you have to have `flake8` installed (`pip install -U flake8`) and `jshint` (`npm install -g jshint`).

peterbe:
What the travis code does is that it executes this:
```
flake8 airmozilla --exclude=*/migrations/*,airmozilla/settings/*.py
```
If you run that in your branch you'll see the same two lines of errors as you get in travis.

blossomica:
@peterbe I did some changes to fix the PEP8 error warnings based on the instructions given. However it seems something is still failing?

airmozilla/main/forms.py
```diff
@@ -47,20 +47,23 @@ class Meta:
...
+            self.fields['recruitmentmessage'].required = False
+            self.fields['recruitmentmessage'].queryset = (
+                RecruitmentMessage.objects.filter(active=True)
+            )
+            self.fields['picture'].widget = GallerySelect(event=self.event)
+            self.fields['picture'].label = (
+                'Select an existing picture from the gallery'
+            )
+        else:
+            del self.fields['call_info']
```
peterbe: So the only thing that should be affected by this new piece of business logic is the `call_info`. 
So leave the other lines as they were and just do:
```python
if not (self.event.is_upcoming() or self.event.is_live()):
    del self.fields['call_info']
```
```

**Comment by blossomica** [comment]
@peterbe - yay! This one works! I wonder why it pyflakes still gives me a "bad substitution" message when I try to commit so I end up having to use ' git commit -n -m ... '

**Comment by peterbe** [comment]
What is the output pyflakes is giving you?

**Comment by blossomica** [comment]
.git/hooks/pre-commit: 7: .git/hooks/pre-commit: Bad substitution

**Comment by peterbe** [comment]
Is line 7 the same as [this line](https://gist.github.com/peterbe/bba51ea999a3a1e36cc5#file-pre-commit-L7) in your system?
What is the output you get when you run: `git diff --cached --name-only --diff-filter=ACM | sort | uniq`

**Comment by blossomica** [comment]
Nothing. No output.

**Comment by peterbe** [comment]
That's because you probably have already done a `git commit -n` so there's nothing left to be committed.

**Comment by blossomica** [comment]
Is that a bad thing?

**Comment by peterbe** [comment]
No, it just means we can't debug that strange message any more. Maybe next PR :)

File: airmozilla/main/templates/main/event.html [hunk 3]
```diff
@@ -188,7 +188,7 @@ <h1 class="entry-title">
         </div>
       {% endif %}
 
-      {% if event.call_info %}
+      {% if event.call_info and event.is_upcoming() or event.is_live() %}
```
peterbe: check your brackets :)

peterbe:
Also, we'll need a test now. You need to create a test that views a live event (with a set `call_info`) and check that that appears and then later disappears on a second view after you've made it archived.

blossomica:
@peterbe - I believe I fixed the bracket issue so now I have to figure out how to write the test. I am looking at main/test_views.py where I think it is the place I am suppose to write the test us. I'll try to find out how to go about it.

blossomica:
@peterbe - I think there is some progress. This is my first attempt so far and it gives me the responses I want when I checked on it. Could you provide some feedback on if it is going well or there is something that is not right and needs fixing?

blossomica:
@peterbe - I did some changes based  on the feedback. You can have a look at it again when you get the chance.

peterbe:
That is really good! That's actually perfect! 

r+

blossomica:
@peterbe - Yay! Ha ha, this is my first test. Thanks for helping me get unstuck. Now i'm waiting for Travis.

peterbe:
Which bug are you going to work on next? Perhaps another one that adds test coverage?

blossomica:
@peterbe All right. I can work on something like that. Do you have any bug about test coverage you suggest I work on from the ones listed ?

peterbe:
This one could be good https://bugzilla.mozilla.org/show_bug.cgi?id=1141306

blossomica:
Okay I'll start working on that.

PR #299, opened 2015-05-08
fixes bug 1141306 - picture_add() lacks a lot of test coverage

blossomica:
@peterbe - Umm, how does this look like like? Any feedback on it?

### DIFF
--- a/airmozilla/manage/tests/views/test_picturegallery.py
+++ b/airmozilla/manage/tests/views/test_picturegallery.py
@@ -242,6 +242,18 @@ class TestPictureGallery(ManageTestCase):
         assert picture.size
         eq_(picture.modified_user, self.user)
 
+    def test_picture_add_with_event(self):
+        event = Event.objects.get(title='Test event')
+        url = reverse('manage:event_edit', args=(event.id,))
+        response = self.client.get(url, {'event': event.id})
+        eq_(response.status_code, 200)
+        ok_(event.title in response.content)
+
+        url = reverse('manage:picture_add')
+        response = self.client.get(url, {'event': event.id})
+        eq_(response.status_code, 200)
+        ok_('?event=%d' % event.id in response.content)
+
     def test_redirect_picture_thumbnail(self):
         with open(self.main_image) as fp:
             picture = Picture.objects.create(


airmozilla/manage/tests/views/test_picturegallery.py
```diff
@@ -242,6 +242,27 @@ def test_picture_add(self):
...
+        event.placeholder_img = event.placeholder_img
+        event.picture = picture
+        event.save()
+        eq_(Picture.objects.all().count(), 1)
+        delete_url = reverse('manage:picture_delete', args=(picture.id,))
+        response = self.client.post(delete_url, data={'event': event.id})
+        eq_(response.status_code, 200)
+        eq_(Picture.objects.filter(event=event).count(), 0)
+        eq_(Picture.objects.all().count(), 0)
+
```
peterbe: This all should be a separate test. There are basically two ways you can add a picture to the gallery. 
1) Just adding a picture. Plain and simple. 
2) Having an event in mind when you're adding a picture. 

For example, click on the event edit of an event somewhere in the event manager. For the picture there should be a link there that says "Pick another" or something. It takes you to the picture gallery management page **but** with the `?event=12345` thing added to the URL. And when you arrive on the gallery page it says "You are here to associate a picture with TITLE OF EVENT".

So you're there to pick a picture. If you want to add a new picture (instead of picking one) we need to remember which event ID it was you were focusing on. That's what this is about. 

So, I would start a new test called something like `test_picture_add_with_event`. 
You do the same `event = Event.objects.get(title='Test event')` and `response = self.client.get(url, {'event': event.id})` but this time you check that the title is in the `response.content` on the form for uploading a picture. 

And you should check that there's a link there on that page back to the picture gallery WITH that `?event=12345` in it.

blossomica:
@peterbe - Is this correct?

airmozilla/manage/tests/views/test_picturegallery.py
```diff
@@ -242,6 +242,16 @@ def test_picture_add(self):
         assert picture.size
         eq_(picture.modified_user, self.user)
 
+    def test_picture_add_with_event(self):
+        event = Event.objects.get(title='Test event')
+        url = reverse('manage:event_edit', kwargs={'id': event.id})
```
peterbe: I usually just do:
```python
url = reverse('manage:event_edit', args=(event.id,))
```
which is shorter.

airmozilla/manage/tests/views/test_picturegallery.py
```diff
@@ -242,6 +242,16 @@ def test_picture_add(self):
...
+        response = self.client.get(url, {'event': event.id})
+        eq_(response.status_code, 200)
+        ok_(event.title in response.content)
+
+        url = reverse('manage:picturegallery') + '?event=%d' % event.id
+        ok_(url in response.content)
+
```
peterbe: So this isn't testing much. I think what you want to test is **going** to the picture add page, but with the `event` parameter on. E.g. 
```python
# add a picture with an event id on it
url = reverse('manage:picture_add', {'event': event.id)
ok_(event.title in response.content)
ok_(event_edit_url in response.content)
```

blossomica:
@peterbe - what do you think of these changes? I feel it got confusing with the example, there is no event.title in the picture_add page from what I see.

peterbe:
You're right! There is no event title on that page. Just a link that says "go back to the Picture Gallery".

r+

blossomica:
@peterbe - I squashed it, so I believe it is finally ready :)

PR #302, opened 2015-05-11
Bug 1151024 - Show correct tootip on starred events - For review only

bugzPDX:
@peterbe I'm stuck on this one. Will you look it over and tell me what I am doing incorrectly?

### DIFF
--- a/airmozilla/main/templates/main/_events.html
+++ b/airmozilla/main/templates/main/_events.html
@@ -20,8 +20,8 @@
     </a>
 
     <a class="star{% if star_on %} star-on{% endif %}" data-id="{{ event.id }}" 
-        data-ttOn="Remove starred event"
-        data-ttOff="Save as a starred event"
+        data-star-on="Remove star"
+        data-star-off="Save as a starred event"
         ></a>
     <p>
       {{ short_desc(event) | safe_html }}
--- a/airmozilla/main/templates/main/_events.html
+++ b/airmozilla/main/templates/main/_events.html
@@ -19,7 +19,10 @@
       {% endif %}
     </a>
 
-    <a class="star{% if star_on %} star-on{% endif %}" data-id="{{ event.id }}" title="Save as a starred event"></a>
+    <a class="star{% if star_on %} star-on{% endif %}" data-id="{{ event.id }}" 
+        data-ttOn="Remove starred event"
+        data-ttOff="Save as a starred event"
+        ></a>
     <p>
       {{ short_desc(event) | safe_html }}
       <a class="go" href="{{ href }}">{{ _('See more') }}</a>
--- a/airmozilla/search/templates/search/home.html
+++ b/airmozilla/search/templates/search/home.html
@@ -92,10 +92,7 @@
 
           <a class="go" href="{{ href }}">{{ _('See more') }}</a>
         </p>
-        <a class="star" data-id="{{ event.id }}"
-            data-ttOn="Remove starred event"
-            data-ttOff="Save as a starred event"
-            ></a>
+        {{ star_tag(event.id)|safe }}
       </div>
     </article>
   {% else %}
--- a/airmozilla/search/templates/search/home.html
+++ b/airmozilla/search/templates/search/home.html
@@ -92,7 +92,10 @@
 
           <a class="go" href="{{ href }}">{{ _('See more') }}</a>
         </p>
-        <a class="star" data-id="{{ event.id }}" title="Save as a starred event"></a>
+        <a class="star" data-id="{{ event.id }}"
+            data-ttOn="Remove starred event"
+            data-ttOff="Save as a starred event"
+            ></a>
       </div>
     </article>
   {% else %}
--- a/airmozilla/starred/helpers.py
+++ b/airmozilla/starred/helpers.py
@@ -1,0 +1,13 @@
+from jingo import register
+
+
+@register.function
+def star_tag(id, extra_classes=None):
+    extra_classes = extra_classes or []
+    if isinstance(extra_classes, basestring):
+        extra_classes = [extra_classes]
+    extra_classes.insert(0, 'star')
+    return '<a class="{0}" data-id="{1}" data-star-on="Remove star" data-star-off="Save as a starred event"></a>'.format(
+        ' '.join(extra_classes),
+        id
+    )
--- a/airmozilla/starred/static/starred/js/star_event.js
+++ b/airmozilla/starred/static/starred/js/star_event.js
@@ -97,11 +97,10 @@ var Stars = (function() {
         setToolTip: function (element) {
             var title;
             var $element = $(element);
-                console.log($element);
             if ($element.hasClass('star-on')) {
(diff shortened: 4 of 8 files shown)


peterbe:
I guess the reason it's not working is because of the case sensitivity. See comments within...

airmozilla/main/templates/main/_events.html
```diff
@@ -19,7 +19,10 @@ <h2 class="entry-title">
       {% endif %}
     </a>
 
-    <a class="star{% if star_on %} star-on{% endif %}" data-id="{{ event.id }}" title="Save as a starred event"></a>
+    <a class="star{% if star_on %} star-on{% endif %}" data-id="{{ event.id }}" 
+        data-ttOn="Remove starred event"
+        data-ttOff="Save as a starred event"
```
peterbe: Change to `data-tton` or `data-star-on`.

airmozilla/search/templates/search/home.html
```diff
@@ -92,7 +92,10 @@ <h5 class="entry-upcoming">
 
           <a class="go" href="{{ href }}">{{ _('See more') }}</a>
         </p>
-        <a class="star" data-id="{{ event.id }}" title="Save as a starred event"></a>
+        <a class="star" data-id="{{ event.id }}"
+            data-ttOn="Remove starred event"
+            data-ttOff="Save as a starred event"
+            ></a>
```
peterbe: This is now the 4th place where we write this down. 

To begin with, I think we should change the wording from "Remove starred event" to "Remove star" (because you're not removing the event, after all). But it means you'd have to do a careful search-and-replace and there's a chance you miss one possibly. 

The only difference is that they have a different extra class. 

The right thing to do would now be to move this piece of HTML into a template helper. I would suggest you create a file called `helpers.py` in the `starred` app and register a function called something like `startag()` which could look something like this:
```python
@register.function
def startag(id, extra_class=''):
    if isinstance(extra_classes, basestring):
       extra_classes = [extra_classes]
    extra_classes.insert(0, 'star')
    return '<a class="{0}" data-id="{1}" data-star-on="Remove..." data-star-off="Save..."></a>'.format(
        ' '.join(extra_classes), 
        id
    )
    
```

airmozilla/starred/static/starred/js/star_event.js
```diff
@@ -99,7 +112,10 @@ $(function() {
 
     $('#content').on('click', 'a.star', function () {
         var id = $(this).data('id');
-        $('a.star[data-id=' + id + ']').toggleClass('star-on');
+        $('a.star[data-id=' + id + ']').each(function(i, element) {
+            $(element).toggleClass('star-on');
+            Stars.setToolTip(element);
+        });
```
peterbe: You can save yourself some jQuery computation by only converting it to a jQuery object once. E.g.
```javascript
$('a.star[data-id=' + id + ']').each(function() {
    var $element = $(this);
    $element.toggleClass('star-on');
    Stars.setToolTip($element); 
});
```
(function `setToolTip` would obviously need to be modified too since it'd receive an `$element` and not `element`)

peterbe: One concern I have is that there's now more stuff in the way of what needs to be fast. The previous line (`$('a.star[data-id=' + id + ']').toggleClass('star-on');`) was good in that it was fast. It doesn't need to do a `.each()` loop thing. Especially since within each loop you have to do that conversion from HTML element to a jQuery element **and** you have to do that `if ($element.hasClass('star-on'))` which is individually fast but when done for 20 elements it adds up. 

Perhaps a much better way would be to re-think the architecture a bit. Perhaps something like:
```javascript
// this is visually important to users on load
$('a.star[data-id=' + id + ']').toggleClass('star-on');  
// these can wait a couple of milliseconds
$('a.star').not('star-on').each(Stars.setTitleOff);  
$('a.star.star-on').each(Stars.setTitleOn);
```

airmozilla/main/templates/main/_events.html
```diff
@@ -19,7 +19,10 @@ <h2 class="entry-title">
       {% endif %}
     </a>
 
-    <a class="star{% if star_on %} star-on{% endif %}" data-id="{{ event.id }}" title="Save as a starred event"></a>
+    <a class="star{% if star_on %} star-on{% endif %}" data-id="{{ event.id }}"
```
bugzPDX: This will also need to use the helper function but I wasn't sure the best way to deal with the if statement.

peterbe: `{{ star_tag(event.id, star_on and 'star-on' or None) }}` might work.

airmozilla/starred/helpers.py
```diff
@@ -0,0 +1,13 @@
...
+@register.function
+def star_tag(id, extra_classes=None):
+    extra_classes = extra_classes or []
+    if isinstance(extra_classes, basestring):
+        extra_classes = [extra_classes]
+    extra_classes.insert(0, 'star')
+    return '<a class="{0}" data-id="{1}" data-star-on="Remove star" data-star-off="Save as a starred event"></a>'.format(
+        ' '.join(extra_classes),
+        id
+    )
```
peterbe: As you might have seen, Travis rejected this because the line was too long. You might have to write it like this:
```python
html = (
    '<a class="{0}" ... '
    'data-star-off="..." '
    '</a>'.format(
        ' '.join(extra_classes),
        id
    )
)
```

peterbe: Also, you should use `from django.utils.safestring import mark_safe` and return a safe string. E.g.
```python
    html = ... (as mentioned above)
    return mark_safe(html)
```
Then you can remove the `| safe` stuff in the templates.

airmozilla/starred/static/starred/js/star_event.js
```diff
@@ -35,6 +35,7 @@ var Stars = (function() {
                 if (index > -1) {
                     $element.addClass('star-on');
                 }
+                Stars.setToolTip(element);
```
peterbe: To avoid an extra `$element = $(element)` inside setToolTip, change this line to `Stars.setToolTip($element)` and change `setToolTip` accordingly.

peterbe:
Looking very promising! However, it concerns me of all all the looping over jQuery elements and then sending them each in to setToolTip which needs it back as a jQuery object. 
A much better approach would be to solve this with a local jQuery plugin. That way you should be able to do things like `$('a.star').setStarToolTip()`. 

A plugin could look something like this:

```javascript
$.fn.setStarToolTip = function() {
    var title;
    if (this.hasClass('star-on')) {
      title = this.data('star-on');
    } else {
      title = this.data('star-off');
   }
   this.attr('title', title);
   return this;
};

var Stars = (function() {
    ...
})();

$(function() {
    ...
});
```

peterbe:
This is also going to be much faster since we're not extracting the original DOM nodes as it loops over the jquery internal arrays.

peterbe:
This got rebased and "embaked" in https://github.com/mozilla/airmozilla/pull/479

PR #310, opened 2015-05-14
fixes bug 1154490 - Need additional event status (eg. processing)

blossomica:
@peterbe - Does this solve "A" from here ? https://bugzilla.mozilla.org/show_bug.cgi?id=1154490 

If yes then you can tell me about "B" as you said, if no then  I am open to feedback.  I saw your wrote "OR" so do we need an "if - else" sort of thing?

### DIFF
--- a/airmozilla/main/models.py
+++ b/airmozilla/main/models.py
@@ -9,6 +9,7 @@ from django.core.cache import cache
 from django.db import models
 from django.dispatch import receiver
 from django.utils import timezone
+from django.db.models import Q
 
 from airmozilla.base.utils import unique_slugify
 from airmozilla.main.fields import EnvironmentField
@@ -213,6 +214,13 @@ class EventManager(models.Manager):
     def scheduled(self):
         return self.get_query_set().filter(status=Event.STATUS_SCHEDULED)
 
+    def scheduled_or_processing(self):
+        return self.get_query_set().filter(
+            Q(status=Event.STATUS_SCHEDULED)
+            |
+            Q(status=Event.STATUS_PROCESSING)
+        )
+
     def approved(self):
         return (
             self.scheduled()
@@ -227,14 +235,15 @@ class EventManager(models.Manager):
         )
 
     def live(self):
-        return self.scheduled().filter(
+        return self.get_query_set().filter(
+            status=Event.STATUS_SCHEDULED,
             archive_time=None,
             start_time__lt=_get_live_time()
         )
 
     def archived(self):
         _now = _get_now()
-        return self.scheduled().filter(
+        return self.scheduled_or_processing().filter(
             archive_time__lt=_now,
             start_time__lt=_now
         )
@@ -255,11 +264,13 @@ class Event(models.Model):
     STATUS_INITIATED = 'initiated'
     STATUS_SCHEDULED = 'scheduled'
     STATUS_PENDING = 'pending'
+    STATUS_PROCESSING = 'processing'
     STATUS_REMOVED = 'removed'
     STATUS_CHOICES = (
         (STATUS_INITIATED, 'Initiated'),
         (STATUS_SCHEDULED, 'Scheduled'),
         (STATUS_PENDING, 'Pending'),
+        (STATUS_PROCESSING, 'Processing'),
         (STATUS_REMOVED, 'Removed')
     )
     status = models.CharField(max_length=20, choices=STATUS_CHOICES,
--- a/airmozilla/main/views.py
+++ b/airmozilla/main/views.py
@@ -340,7 +340,12 @@ class EventView(View):
             return self.cant_view_event(event, request)
 
         warning = None
-        if event.status not in (Event.STATUS_SCHEDULED, Event.STATUS_PENDING):
+        ok_statuses = (
+            Event.STATUS_SCHEDULED,
+            Event.STATUS_PENDING,
+            Event.STATUS_PROCESSING,
+            )
+        if event.status not in ok_statuses:
             if not request.user.is_superuser:
                 self.template_name = 'main/event_not_scheduled.html'
             else:
@@ -971,7 +976,7 @@ def events_calendar_ical(request, privacy=None):
     cal = vobject.iCalendar()
 
     now = timezone.now()
-    base_qs = Event.objects.scheduled()
+    base_qs = Event.objects.scheduled_or_processing()
     if privacy == 'public':
         base_qs = base_qs.approved().filter(
             privacy=Event.PRIVACY_PUBLIC
@@ -1071,7 +1076,7 @@ class EventsFeed(Feed):
     def items(self):
         now = timezone.now()
         qs = (
-            Event.objects.scheduled()
+            Event.objects.scheduled_or_processing()
             .filter(start_time__lt=now,
                     channels=self._channel)
             .order_by('-start_time')
@@ -1223,7 +1228,7 @@ def calendar_data(request):
 
     privacy_filter = {}
     privacy_exclude = {}
-    events = Event.objects.scheduled()
+    events = Event.objects.scheduled_or_processing()
     if request.user.is_active:
         if is_contributor(request.user):
             privacy_exclude = {'privacy': Event.PRIVACY_COMPANY}
--- a/airmozilla/manage/archiver.py
+++ b/airmozilla/manage/archiver.py
@@ -6,6 +6,7 @@ from django.core.mail import EmailMessage
 from django.core.cache import cache
 from django.utils import timezone
 from django.contrib.sites.models import Site
+from django.db.models import Q
 
 from funfactory.urlresolvers import reverse
 
@@ -18,7 +19,9 @@ from .vidly import query
 def auto_archive(verbose=False):
     events = (
         Event.objects
-        .filter(status=Event.STATUS_PENDING,
+        .filter(Q(status=Event.STATUS_PENDING)
+                |
+                Q(status=Event.STATUS_PROCESSING),
                 archive_time__isnull=True,
                 template__name__contains='Vid.ly')
     )
@@ -29,7 +32,7 @@ def auto_archive(verbose=False):
 
     if verbose:  # pragma: no cover
         if any:
-            print "No pending events."
+            print "No processing or pending events."
         else:
             print "No events to archive."
 
@@ -96,7 +99,8 @@ def archive(event, swallow_email_exceptions=False, verbose=False):
                 )
                 if verbose:  # pragma: no cover
                     print (
-                        'Unable to archive pending event "%s" with tag %s' % (
+                        'Unable to archive event "%s" with tag %s'
+                        % (
                             event,
                             tag
                         )
@@ -145,7 +149,7 @@ def build_absolute_url(uri):
 
 
 def email_about_archiver_error(event, tag):
-    subject = "Unable to archive pending event with tag %s" % tag
+    subject = "Unable to archive event with tag %s" % tag
     message = (
         'When trying to archive the "%s" event we had an error from Vid.ly.\n'
         '\n'
--- a/airmozilla/manage/tests/test_archiver.py
+++ b/airmozilla/manage/tests/test_archiver.py
@@ -119,7 +119,7 @@ class ArchiverTestCase(TestCase):
 
         sent_email = mail.outbox[-1]
         eq_(sent_email.to, [x[1] for x in settings.ADMINS])
-        ok_('Unable to archive pending event' in sent_email.subject)
+        ok_('Unable to archive event' in sent_email.subject)
         ok_('abc123' in sent_email.subject)
         ok_(reverse('manage:event_edit', args=(event.pk,)) in sent_email.body)
 
--- a/airmozilla/manage/views/events.py
+++ b/airmozilla/manage/views/events.py
@@ -635,7 +635,7 @@ def event_stop_live(request, id):
     """Convenient thing that changes the status and redirects you to
     go and upload a file."""
     event = get_object_or_404(Event, id=id)
-    event.status = Event.STATUS_PENDING
+    event.status = Event.STATUS_PROCESSING
     event.save()
 
     return redirect('manage:event_upload', event.pk)
@@ -1058,7 +1058,12 @@ def event_archive(request, id):
                 event.has_vidly_template() and
                 not has_successful_vidly_submission(event)
             ):
-                event.status = Event.STATUS_PENDING
+                # some events go from live -> to transcoding -> to archived
+                # and others go right away from transcoding -> to archived
+                # this IF makes sure the event is not currently transcoding
+                # before its status changes to PENDING
+                if event.status != Event.STATUS_PROCESSING:
+                    event.status = Event.STATUS_PENDING
             else:
                 event.status = Event.STATUS_SCHEDULED
                 now = (
--- a/airmozilla/search/views.py
+++ b/airmozilla/search/views.py
@@ -42,7 +42,7 @@ def home(request):
         context['q'] = form.cleaned_data['q']
         privacy_filter = {}
         privacy_exclude = {}
-        qs = Event.objects.scheduled()
+        qs = Event.objects.scheduled_or_processing()
         if request.user.is_active:
             if is_contributor(request.user):
                 privacy_exclude = {'privacy': Event.PRIVACY_COMPANY}
--- a/airmozilla/main/templates/main/event.html
+++ b/airmozilla/main/templates/main/event.html
@@ -62,6 +62,11 @@
         <p class="pending">
         Event is currently in a pending state meaning it will soon be made available.
         </p>
+      {% elif processing %}
+        <p>
+        This video is currently being transcoded; that will take some time. 
+        Feel free to check again on its status after a while.
+        </p>
       {% else %}
         {% if video %}
           {{ video|safe }}
--- a/airmozilla/manage/autocompeter.py
+++ b/airmozilla/manage/autocompeter.py
@@ -63,10 +63,10 @@ def update(
             median_hits = sorted(values)[len(values) / 2]
         else:
             median_hits = 0
-        events = Event.objects.scheduled()
+        events = Event.objects.scheduled_or_processing()
     else:
         events = (
-            Event.objects.scheduled()
+            Event.objects.scheduled_or_processing()
             .filter(modified__gte=now-since)[:max_]
         )
         if events:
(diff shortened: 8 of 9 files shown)


blossomica:
I tried to work on B as well after checking out several files and trying to understand what's going on, so the 2nd PR is my attempt on B though I am not sure if that is on the right track.

airmozilla/manage/archiver.py
```diff
@@ -18,7 +18,7 @@
 def auto_archive(verbose=False):
     events = (
         Event.objects
-        .filter(status=Event.STATUS_PENDING,
+        .filter(status=Event.STATUS_PROCESSING,
```
peterbe: No, this needs to be:
```python
.filter(Q(status=Event.STATUS_PENDING) | Q(status=Event.STATUS_PROCESSING))
```
You can find examples of using `Q` in other places of the code. 
Basically, what this would do is make it an OR statement. 

Some events (e.g. uploaded pre-recorded files) will go straight into PENDING and those that *were* live will go into PROCESSING status.

airmozilla/manage/archiver.py
```diff
@@ -29,7 +29,7 @@ def auto_archive(verbose=False):
 
     if verbose:  # pragma: no cover
         if any:
-            print "No pending events."
+            print "No processing events."
```
peterbe: "No process or pending events" maybe

blossomica:
@peterbe: I did some changed based on the feedback.

airmozilla/main/models.py
```diff
@@ -213,6 +213,9 @@ def get_queryset(self):
     def scheduled(self):
         return self.get_query_set().filter(status=Event.STATUS_SCHEDULED)
 
+    def processing(self):
+        return self.get_query_set().filter(status=Event.STATUS_PROCESSING)
+
```
peterbe: Are you sure we need this extra method?

airmozilla/manage/views/events.py
```diff
@@ -635,7 +635,7 @@ def event_stop_live(request, id):
     """Convenient thing that changes the status and redirects you to
     go and upload a file."""
     event = get_object_or_404(Event, id=id)
-    event.status = Event.STATUS_PENDING
+    event.status = Event.STATUS_PROCESSING
```
peterbe: Excellent! This is the critical change.

peterbe:
It looks really promising! Unfortunately, I think I have to download it and play with a bit and mess around with submitting some videos. It might take some time. I don't want you to be blocked on me. I suggest you perhaps start already looking at another bug.

blossomica:
@peterbe: Okay. So should I remove this last extra method or you will take over this bug entirely so I should look for another one?

peterbe:
I will not take over the bug. I just need to test it a bit more extensively. 
For example, have you looked at what happens when you try to view an event whose status==PENDING (or PROCESSING). Basically, the same thing should happen.

blossomica:
@peterbe: No I haven't tested it at all because I wasn't able to figure out how to get a video to the transcribing stage for me to reproduce that bug so no I haven't tried that at all, and all right i'll try to work on some other bug till this gets tested.

airmozilla/manage/archiver.py
```diff
@@ -145,7 +150,7 @@ def build_absolute_url(uri):
 
 
 def email_about_archiver_error(event, tag):
-    subject = "Unable to archive pending event with tag %s" % tag
+    subject = "Unable to archive process or pending event with tag %s" % tag
```
peterbe: Here and two places above, I think it should be "...processing or pending event..."

blossomica:
@peterbe - I fixed those.

peterbe:
This all looks great! But it's only part B. I think, the only thing you're going to need to change is to [this crucial line](https://github.com/mozilla/airmozilla/blob/5490131307b954fa073e1180166f3ffc618a6dc9/airmozilla/main/models.py#L214) so that it does one of those `Q(status=Event.STATUS_SCHEDULED) | Q(status=Event.STATUS_PROCESSING)`.

That's the aspect that I'd like to do some manual testing of.

blossomica:
@peterbe: I added it. Let me know if there is anything I should fix.

peterbe:
Cool! I'll play with it now. For realz.

blossomica:
LOL sounds exciting!

peterbe:
This happened when I tried to view it when it was in Processing status
screenshot 2015-05-19 10 37 03

blossomica:
Okay. So I guess here is where this comes into play 

> It would be better to be able to see the event at least. And a link to it. On that page we'll say "Hey,  thanks for stopping by. We're currently transcoding this. Please be patient and check back in a couple > of minutes".

When you say "see the event" what do you really mean, and do you want that red warning (I assume that will be in some JavaScript file) to write "Hey thanks... a couple of minutes" ?

peterbe:
Actually, doing the automatic conversion from live to processing didn't work. I think what happens is that when you click the red "Stop live event" it executes [this function](https://github.com/blossomica/airmozilla/blob/3db81230c5c5cacd63ef80cdac3833c6d67f167f/airmozilla/manage/views/events.py#L634-L639) first and that just makes the event no longer Scheduled. However, when you go through that, the next thing that happens is that it executes [this function](https://github.com/blossomica/airmozilla/blob/3db81230c5c5cacd63ef80cdac3833c6d67f167f/airmozilla/manage/views/events.py#L1057-L1061).

So the first function simply makes it NOT Scheduled. But the second one really makes sure it's Pending and that's wrong. That last function should set it to Processing because that's more true. However, it should ONLY do that if it was previously Processing. E.g. Change it to:
```
if event.status != Event.STATUS_PROCESSING:
    event.status = Event.STATUS_PENDING
```
Does that make sense?

peterbe:
> When you say "see the event" what do you really mean, and do you want that red warning

E.g. A link to it should be made available. For example from the Calendar page, or the home page or the search results.

peterbe:
I don't think any new javascript is going to be needed. 
To start with, this needs to change: https://github.com/blossomica/airmozilla/blob/3db81230c5c5cacd63ef80cdac3833c6d67f167f/airmozilla/main/views.py#L343

peterbe:
Another thing to look into is that `myevent.is_live()` should never be true if `myevent.status==Event.STATUS_PROCESSING`.

At the moment, the event that I put into status=Processing (manually) still appears on the home page as "Streaming Live Now" :)
screenshot 2015-05-19 10 52 54

Look into https://github.com/blossomica/airmozilla/blob/3db81230c5c5cacd63ef80cdac3833c6d67f167f/airmozilla/main/models.py#L232-L236 and you might need to stop using `self.scheduled()` and replace it with `self.get_query_set().filter(Q(status=Event.STATUS_SCHEDULED)`.

Lastly, I'm wondering. Is it maybe a bad idea to abuse the [`scheduled()` method](https://github.com/blossomica/airmozilla/blob/3db81230c5c5cacd63ef80cdac3833c6d67f167f/airmozilla/main/models.py#L214) to mean something different than what the name of the method is. 
I know this was my original suggestion but I'm having second thoughts. 

Basically, we want links to events to be visible to the user on the home page, channel browsing, search results, calendar, iCal, RSS/Atom feeds if `(status=Processing OR status=Scheduled)`. 

Perhaps we instead just look for places where we do `Event.objects.scheduled()` and replace that with `Event.objects.scheduled_or_processing()`. 

What do you think? Were we too lazy when we took advantage of that `Event.objects.scheduled()`?

peterbe:
Another crucial change we need to make is [here](https://github.com/blossomica/airmozilla/blob/3db81230c5c5cacd63ef80cdac3833c6d67f167f/airmozilla/main/templates/main/event.html#L61-L65).

If an event is current in the state of Processing it'll end up in that `{% else %}` block which means it tries to actually display the video which it shouldn't possibly do if Vid.ly hasn't actually transcoded it yet. 

So you'll need to add another...:

```
{% elif processing %}
<p>Video is being transcoded as we speak. It's going to take a little while. </p>
{% else %}
...
```
(please feel free to think up a better more user-friendly sentence to put there)

blossomica:
I think that it might look lazy in the long run because overtime if someone else studies the code that part might be confusing and they might wonder why we used a method to do something different than it was intended for. So for a temporal fix it looks okay but when I look at its overtime effects I think it will benefit other potential future contributors I think it will be nice if that part could be clearer. And sure, I'll think of a more user-friendly sentence when I get to that step.

peterbe:
What's also really exciting is that we can use that state there on that page to do two possibly cool things:

1) Use some javascript that asks the server who long we think it's going to until it's transcoded. 
We can almost know this now because we're developing a formula for working out how long it takes to transcode videos. It's roughly 1.8 * duration_of_original. So if the uploaded video is 200s, the transcoding is going to take about 6 minutes. And we know when we submitted it to Vid.ly (see the `VidlySubmission` class) so we if we submitted it 2 minutes ago we can say "We estimate it will be available in about 4 minutes"

2) We can set up a javascript interval that constantly asks the server "Is the video ready yet?" And if the server says Yes, we reload the page. Something like this:
```
{% if processing %}
<script>
setInterval(function() {
    $.getJSON('readyyet?', function(r) {
       if (r.yes) document.location.reload();
    });
}, 5000);
</script>
{% endif %}
```
Actually the details are likely to be very different and dealt with in a fresh new bug.

peterbe:
> ... someone else studies the code that part might be confusing and they might wonder...

Ok. That settles it then. We both have the same feeling. Go ahead and change back the old `scheduled()` method and add a new one called `scheduled_or_processing()`.

I love the fact that we've finally found a bug that is a bit more challenging for you.

blossomica:
Haha yes this bug gets harder it seems. It escalated very fast.

peterbe:
> Haha yes this bug gets harder it seems.

But this is a very exciting feature. A lot of people have been fooled by events suddenly disappearing whilst it's being transcoded. They think they made some mistake in their calendaring when it's just that the event is undergoing a transformation.

blossomica:
@peterbe - I attempted to fix most of the things that went wrong...I am not sure if it is okay so far so you can let me know if something is wrong there.

I attempted to fix everything except this place below where I need some further clarification on:
> Perhaps we instead just look for places where we do Event.objects.scheduled() and replace that with > Event.objects.scheduled_or_processing().

I see `Event.objects.scheduled()` on several files for example `airmozilla/new/eventemails.py ` does that really have to change everywhere. I am a bit confused on which places it needs to change to ` Event.objects.scheduled_or_processing()` and which places it doesn't need to.

peterbe:
> I am a bit confused on which places it needs to change to `Event.objects.scheduled_or_processing()` and which places it doesn't need to. 

Well, I guess your guess is almost as good as mine. 
There are several ways to list events:
* The home page
* The sidebar (appears on home page and individual events (except then they're not on the "side"))
* Search results
* Calendar
* Roku (definitely shouldn't include the processing ones)
* iCal feeds
* RSS/Atom feeds
* `eventemails.py` (should not use processing ones)

So basically all but the Roku stuff and the `eventemails.py` should switch to using `scheduled_or_processing()` instead.

blossomica:
@peterbe - I fixed those little bugs, I had an issue yesterday and I couldn't get access to the IRC which is why I sent an email, but I sorted out so no worries. I am super glad to be able to finish this huge bug it seems to be towards its completion :)

peterbe:
So with the last commit you started switching over to using `.scheduled_or_processing()` in some places. But aren't there more places next where we need to change too? E.g. the sidebar (see the context_processors.py file)

blossomica:
@peterbe - I checked it but I don't really see where the change is needed at? I don't see the .sheduled() being used there.

peterbe:
That's the problem with being such as fast learner as you are that when you attack harder bugs it requires a bit more digging and debugging :)

blossomica:
@peterbe I tried to rebase against master but I am not sure what happened here.

airmozilla/manage/tests/test_archiver.py
```diff
@@ -119,7 +119,8 @@ def mocked_urlopen(request):
 
         sent_email = mail.outbox[-1]
         eq_(sent_email.to, [x[1] for x in settings.ADMINS])
-        ok_('Unable to archive pending event' in sent_email.subject)
+        ok_('Unable to archive processing or pending event'
```
peterbe: Let's just get rid of these and say "Unable to archive event". Which state it is/was in isn't useful.

airmozilla/search/views.py
```diff
@@ -42,7 +42,7 @@ def home(request):
         context['q'] = form.cleaned_data['q']
         privacy_filter = {}
         privacy_exclude = {}
-        qs = Event.objects.scheduled()
+        qs = Event.objects.scheduled_or_processing()
```
peterbe: Great!

peterbe:
Now that you've rebased, you should be able edit [this line](https://github.com/blossomica/airmozilla/blob/bug1154490-Need-additional-event-status-eg-processing/airmozilla/main/views.py#L1229) as well.

peterbe:
So another crucial place to get busy with is [this line](https://github.com/blossomica/airmozilla/blob/bug1154490-Need-additional-event-status-eg-processing/airmozilla/main/views.py#L90).

The root query for the home page is archived events. That query, [as you can see here](https://github.com/blossomica/airmozilla/blob/bug1154490-Need-additional-event-status-eg-processing/airmozilla/main/models.py) depends on `scheduled()`.

One option would be to simply change [this line](https://github.com/blossomica/airmozilla/blob/bug1154490-Need-additional-event-status-eg-processing/airmozilla/main/models.py#L243) to be:
```python
    def archived(self):
        _now = _get_now()
        return self.scheduled_or_processing().filter(
            archive_time__lt=_now,
            start_time__lt=_now
        )
```
But isn't that what Roku depends on? 
But Roku doesn't depend on `.archived()`. It just uses [`.scheduled()`](https://github.com/blossomica/airmozilla/blob/bug1154490-Need-additional-event-status-eg-processing/airmozilla/roku/views.py#L30) and that should stay like that.

blossomica:
@peterbe I didn't get what I am suppose to do in the last bit, are you trying to say .archived() should be .scheduled() instead?

peterbe:
First of all, I think you can figure out why the test broke. It seems fairly straight forward. 

Yes, I think you should change how the `archived()` method works. We'll scrutinize that the Roku stuff works later.

blossomica:
@peterbe haha yes I just  figured that out. I totally forgot of those prints existed. Sorry for that. 
I changed the archived() method. So if it is okay, you can tell me about Roku.

peterbe:
By the way, next time you rebase, you'll pick up the change that will hopefully fix those sporadic errors we've been getting with tests.

blossomica:
@peterbe - I am running the test on my system and so far I am getting the same twitter error. I'll upload the changes and see how it goes.

peterbe:
I think the reason if failed again was because https://github.com/mozilla/airmozilla/commit/76d905fcc65aab3d4749ff6505b6ce197e526db7 **hadn't** landed yet. You're too fast! :)
Do a rebase again.

blossomica:
@peterbe - Haha, that sounds a bit funny. I just rebased it, lets see how it goes now.

airmozilla/main/models.py
```diff
@@ -213,6 +214,11 @@ def get_queryset(self):
     def scheduled(self):
         return self.get_query_set().filter(status=Event.STATUS_SCHEDULED)
 
+    def scheduled_or_processing(self):
+        return self.get_query_set().filter(Q(status=Event.STATUS_SCHEDULED)
+                                           |
+                                           Q(status=Event.STATUS_PROCESSING))
+
```
peterbe: I know it's petty but it actually matters. Can you reformat the indentation so it's like this instead:
```python
return self.get_query_set().filter(
    Q(status=Event.STATUS_SCHEDULED)
    |
    Q(status=Event.STATUS_PROCESSING)
)
```

The reason why it matters is because consistency is important. What's also important is to keep patches as small as possible. For example, if you wrote this:
```python
names = ('Peter',
         'Gloria',
         'Lisa',
         'Erica',
         'Richard',
)
```
and some day you decide to change the variable from `names` to `first_names` for some good reason. Then it would end up misaligned like this:
```python
first_names = ('Peter',
         'Gloria',
         'Lisa',
         'Erica',
         'Richard',
)
```
You're only intending on changing the variable name. Not the list/tuple. The list of names now look odd you think so you decide to change it. All of a sudden your diff becomes that of 5 lines. Perhaps that could cause a git conflict with someone elses changes even though all you changed was indentation. Also a code reviewing might see that change and have to study the difference very careful and think to herself "Hmm... Why did all of these names change? Let me look through each one."
A much better approach is this:
```python
names = (
    'Peter',
    'Gloria',
    'Lisa',
    'Erica',
    'Richard',
)
``` 
Now, the length of the variable `names` is independent of the list of names. The code is thus more "sustainable".

airmozilla/main/models.py
```diff
@@ -227,14 +233,14 @@ def upcoming(self):
...
 
     def live(self):
-        return self.scheduled().filter(
-            archive_time=None,
-            start_time__lt=_get_live_time()
-        )
+        return self.get_query_set().filter(status=Event.STATUS_SCHEDULED,
+                                           archive_time=None,
+                                           start_time__lt=_get_live_time()
+                                           )
```
peterbe: Same thing about the indentation here please.

airmozilla/main/views.py
```diff
@@ -340,7 +340,12 @@ def get(self, request, slug):
             return self.cant_view_event(event, request)
 
         warning = None
-        if event.status not in (Event.STATUS_SCHEDULED, Event.STATUS_PENDING):
+        ok_statuses = (
+            Event.STATUS_SCHEDULED,
+            Event.STATUS_PENDING,
+            Event.STATUS_PROCESSING
```
peterbe: Always end with a comma. This is not true for JSON or for some Javascript linters. 
The reason you want to always end with a comma is that when you later add 1 more item to that tuple, it only becomes one line of change. If you also have to add a new comma to the previous line, it becomes two changed lines. 

Also, by always ending on a comma, when you some day do add another item there's a slower chance of accidentally causing a syntaxerror. 

One place where this rule is not true is when defining parameters to a function. E.g 
```python
def my_amazing_python_function(
    self, 
    argument_number_1,
    argument_number_2,
    argument_number_3
):
    # here is the body of the code
```

peterbe:
r+ with some style changes with I trust you can fix before you do a final rebase. 

It's not impossible that we forgot some place (of replacing `.scheduled()` with `.scheduled_or_processing()`) but I think it's better to move forward and try out what this gives us and if we spot something we'll simply correct it that day.

blossomica:
@peterbe - I fixed the things you mentioned :)

peterbe:
Again you continue to impress me!

blossomica:
@peterbe - Thanks :)

PR #312, opened 2015-05-15
fixes bug 1160664 - Change 'start time' notation depending on video

blossomica:
@peterbe - I tried to fix this bug so you can check it when you get the chance.

PS. It seems something went wrong here, so I have to figure that out how to fix the issue with raven that it has.

### DIFF
--- a/airmozilla/main/templates/main/event.html
+++ b/airmozilla/main/templates/main/event.html
@@ -101,7 +101,7 @@
              &mdash;
              {{ event.location_time.strftime("%I:%M%p") }}
               {% if not event.location.name.startswith('Cyberspace') and not
-                 event.location.name.startswith(settings.DEFAULT_PRERECORDED_LOCATION[0]) %}
+                 event.is_prerecorded() %}
                  in {{ event.location.name }}
               {% endif %}
             {% endif %}
@@ -121,7 +121,11 @@
           <label for="tab-details">Details</label>
           <div class="content">
             <p>
-              Start time: {{ event.start_time | js_date }}
+              {% if event.is_prerecorded() %}
+                  Uploaded: {{ event.start_time | js_date }}
+              {% else %}
+                  Start time: {{ event.start_time | js_date }}
+              {% endif %}
               {% if event.is_upcoming() %}
                 {% if event.location %}
                   &mdash;
@@ -131,7 +135,7 @@
               <br>
 
               {% if event.location and not event.location.name.startswith('Cyberspace')
-                 and not event.location.name.startswith(settings.DEFAULT_PRERECORDED_LOCATION[0]) %}
+                 and not event.is_prerecorded() %}
                 {{ _('Location') }}: {{ event.location.name }}<br>
               {% endif %}
 
--- a/airmozilla/main/models.py
+++ b/airmozilla/main/models.py
@@ -367,6 +367,14 @@ class Event(models.Model):
                     return True
         return False
 
+    def is_prerecorded(self):
+        if self.location:
+            if self.location.name.startswith(
+               settings.DEFAULT_PRERECORDED_LOCATION[0]):
+                return True
+            return False
+        return True
+
     def has_vidly_template(self):
         return self.template and 'Vid.ly' in self.template.name
 


blossomica:
@peterbe - It seems I fixed the previous error I had on this bug. So you can have a look at it when you get the chance.

peterbe:
This patch accidentally deletes a whole file :)

blossomica:
How? Is that what " OK (Skip = 2)" means in the test results?

airmozilla/main/templates/main/event.html
```diff
@@ -121,7 +121,13 @@ <h1 class="entry-title">
...
           <div class="content">
             <p>
-              Start time: {{ event.start_time | js_date }}
+              {% if event.location %}
+                  {% if event.location.name.startswith(settings.DEFAULT_PRERECORDED_LOCATION[0]) %}
+                      Uploaded: {{ event.start_time | js_date }}
+                  {% else %}
+                      Start time: {{ event.start_time | js_date }}
+                  {% endif %}
+              {% endif %}
```
peterbe: There is another situation we need to think about. When you upload a new video, using the new uploader it'll exist in the database **without** a location. E.g. `event.location==None`. But it might still have a `start_time`.

Also, this template code is starting to get complicated. I think you should add a new method to the `Event` model class. Call it "is_prerecorded" and it should look something like this:
```python
def is_prerecorded(self):
    if self.location:
        if self.location.name.startswith(settings.DEFAULT_PRE....:
            return True
        return False
    return True
```
If you could do that please, can you also review other places where the `settings.DEFAULT_PRERECORDED_LOCATION` is being used and if we can do some refactoring there.

peterbe:
The `" OK (Skip = 2)"` happens when you run the whole test suite with `RUN_SELENIUM_TESTS = False`.

On travis we make sure to run it with  `RUN_SELENIUM_TESTS = True` as you can see [here](https://github.com/mozilla/airmozilla/blob/master/bin/travis/setup.sh#L27).

blossomica:
@peterbe - I did the changes in the Event model but when I try to refactor the other places `settings.DEFAULT_PRERECORDED_LOCATION` is used I get some errors from main/tests/test_views.py in the some of the ok_() methods that there is no event.location.name which I suspect might relate to what you said that some uploaded events have no location. When I leave it as is there is no error but when I substitute it with the `event.is_prerecorded()` method it gives me those errors.

peterbe:
Your PR here still has the whole of `local.py-dist` entirely removed. You might need to add it back so when compared to master it doesn't delete that file :)

blossomica:
@peterbe - Yes I deleted that file because I noticed I have two local.py files and I thought that file was similar to the other one without the edits I had done, but i'll add it back, and yes that if - else doesn't make sense which is why I got stuck...I was just trying to get it compile correctly so I can get some feedback on it but it is a bad practice :-/ I'll try to upload the broken code instead next time.

blossomica:
@peterbe - I fixed this and added the local.py-dist, though I have to mention that in my environment I have a "streaming live video" which appears at home and it lists "Start time: 2 months ago" when I click it thought and the event page shows up, it lists in the details tab "Uploaded: 2 months ago" . So this looks really strange I hope this is happening only on my side but then again I still find that strange.

peterbe:
Hmm... So you have a live event that doesn't have have a location? That could just be due to using the fake data script. I don't think that makes sense "in the real world".

blossomica:
Yes it also has no location, but the thing I find strange is the way it says  "Start time: 2 months ago"  and "Streaming live now", while when I click on it it says "Uploaded: 2 months ago". It might be the fake data script as you stated.

peterbe:
You're touching on a weak part of Air Mozilla. Remember, Air Mozilla started out as a live-streams-only thing. Everything was going to be live at some point. It's only in recent years where we have the ability to upload prerecorded video files. That's why we call it "Event" and not "Video". 

Normally, when an event is live, it makes sense to say "Start time: ..." because you might be 10 minutes late and then it makes perfect sense to say "Start time: 10 minutes ago". 

Generally, the producers of Air Mozilla manually take care of stopping live events and turning them into archived events (by making it pending whilst uploading the recording, which is what you're working on changing). But you don't bother with your fake data. For me, with my local development database, on my homepage I have 10+ events that are live and they all say crazy things like "Start time: 6 months ago" etc. 

So, what's the state of this PR now? Do you know what needs to be done or is all a blur?

blossomica:
Oh all right. I didn't know it was initially made only for live events. That makes sense now.
I think this PR is ready to be re-based. What do you think?

peterbe:
r+ but you still need to put back the `local.py-dist`. If you can put that back in and rebase I'll merge.

blossomica:
@peterbe - I rebased but I am not sure why local.py-dist does not appear in the committed files. I did add it.

peterbe:
Another win for @blossomica !

blossomica:
Thanks. I can finally delete that branch and focus on the other bug without the switches.

PR #315, opened 2015-05-20
fixes bug 1164936 - Removed events should not appear in the upcoming …

blossomica:
…assignments list.

@peterbe - I worked on this bug too.

### DIFF
--- a/airmozilla/manage/views/events.py
+++ b/airmozilla/manage/views/events.py
@@ -1345,7 +1345,11 @@ def event_assignments(request):
 
     events = []
     now = timezone.now()
-    qs = Event.objects.filter(start_time__gte=now)
+    qs = (
+        Event.objects
+        .exclude(status=Event.STATUS_REMOVED)
+        .filter(start_time__gte=now)
+    )
     for event in qs.order_by('start_time'):
         try:
             assignment = EventAssignment.objects.get(event=event)
--- a/airmozilla/manage/tests/views/test_eventassignment.py
+++ b/airmozilla/manage/tests/views/test_eventassignment.py
@@ -74,9 +74,9 @@ class TestEventAssignment(ManageTestCase):
 
         # add another, also upcoming, event and make sure it's
         # working for events that have no assignments too
-        Event.objects.create(
+        event1 = Event.objects.create(
             title='Other title',
-            start_time=event.start_time
+            start_time=event.start_time,
         )
         # and add a location to the assignment too
         barcelona = Location.objects.create(name='Barcelona')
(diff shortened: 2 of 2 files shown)


airmozilla/manage/templates/manage/event_assignments.html
```diff
@@ -54,6 +54,7 @@
     <tbody>
     {% for event in events %}
       <tr>
+       {% if not event.is_removed() %}
```
peterbe: This is not the right place to do the filtering. You should do that deeper down. Like here maybe: https://github.com/blossomica/airmozilla/blob/bug1164936-Removed-events-should-not-appear-in-the-upcoming-assignments-list/airmozilla/manage/views/events.py#L1348

blossomica:
@peterbe : I did some edits. Hopefully the last PR is the right one.

peterbe:
No, all you need to do is change this line:
```python
qs = Event.objects.filter(start_time__gte=now)
```
to this:
```python
qs = Event.objects.scheduled().filter(start_time__gte=now)
```

peterbe:
I know it might feel frustrating now but once you see it you'll see how obvious it is. You just need to take a tiny step back and get the idea. It's hard when so much is going on.

blossomica:
@peterbe - Okay. I fixed that. The  other ones worked properly too so sometimes it is hard to know which method to use.

airmozilla/manage/views/events.py
```diff
@@ -1345,7 +1345,7 @@ def __init__(self, user, count):
 
     events = []
     now = timezone.now()
-    qs = Event.objects.filter(start_time__gte=now)
+    qs = Event.objects.scheduled().filter(start_time__gte=now)
```
peterbe: No, we actually just want to exclude removed ones. I would use:
```python
qs = (
    Event.objects
    .exclude(status=Event.STATUS_REMOVED)
    .filter(start_time__gte=now)
)
```

blossomica:
@peterbe okay I fixed this. I did try something similar but I placed .exclude after .filter and it was giving me errors so that's probably a syntax mistake I did and it makes sense in some way on why I got the error that I did. Thanks for the input :)

blossomica:
@peterbe - After second thoughts. This method doesn't work either sadly.

peterbe:
The tests fail because of something that isn't your fault. See https://bugzilla.mozilla.org/show_bug.cgi?id=1167211

I'm going to investigate why that might be.

peterbe:
So, another thing you need to do is prove this change with a test. Look at these lines:
https://github.com/blossomica/airmozilla/blob/bug1164936-Removed-events-should-not-appear-in-the-upcoming-assignments-list/airmozilla/manage/tests/views/test_eventassignment.py#L77-L80

It creates another event and on line 88 it asserts that it's in the `response.content`.

Change that event to `event.status = Event.STATUS_REMOVED` then do what's done on line 86 again and this time assert that 'Other title' is NOT in the `response.content`.

You'll need to change line 77 to something like:
```
other_event = Event.objects.create(....
```
So you can change its status later.

blossomica:
@peterbe - Finally I think this resolved? Let me know if it is time for squashing.

peterbe:
I'll do some manual testing too.

blossomica:
All right, let me know how it goes. This PR is okay in fact. There is another one (the one with processing/scheduled) that gives me the tweets error.  https://github.com/mozilla/airmozilla/pull/310

peterbe:
r+

blossomica:
@peterbe  - Done

PR #334, opened 2015-06-02
Emphasised need for memcached

Nolski:
This seemed pretty easy to miss but since the wording was there I think  it might be a good idea to emphasize this text.

### DIFF
--- a/README.md
+++ b/README.md
@@ -148,7 +148,7 @@ For security, you need to enter something into the `SECRET_KEY`.
 SECRET_KEY = 'somethingnotempty'
 ```
 
-By default, you need a `Memcache` server up and running. The connection
+**By default, you need a `Memcache` server up and running.** The connection
 settings for that is not entered by default. So if you have a Memcache running
 on the default port you need to enter it for the `LOCATION` setting like this:
 ```


README.md
```diff
@@ -148,7 +148,7 @@ For security, you need to enter something into the `SECRET_KEY`.
 SECRET_KEY = 'somethingnotempty'
 ```
 
-By default, you need a `Memcache` server up and running. The connection
+*By default, you need a `Memcache` server up and running.* The connection
```
peterbe: Italic text is for emphasis on a specific word like "The cat was in the *house* not in the barn". 
Use bold for things that are supposed to stand out.

Nolski: For code review changes, would you rather I add an additional commit with the fix or amend my previous commit?

Nolski:
updated to be bold.

README.md
peterbe: Keep adding commits until you get an r+ comment. Then, when you have one of those, go ahead and rebase your PR down to 1 neat messaged commit and force push that into your branch and once the travis tests have tested one last time, I merge.

PR #340, opened 2015-06-02
fixes bug 1170727 - Removing Response class duplication in tests

a-buck:
As Peter Bengtsson advised to do on [bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=1170727), I moved the Response class that was duplicated across multiple test classes into airmozilla.base.tests.testbase.

Response was implemented slightly differently in airmozilla/manage/tests/test_autocompleter vs the rest. The `json()` method returned `json.loads(self.content)` instead of simply `self.content` like the Response class in the other test classes did.

Therefore I subclassed Response from airmozilla.base.tests.testbase in test_autocompleter with the slightly different `json()` method and called it `JsonLoadingResponse` since it is loading the json instead of just returning the content of the response.

Please let me know if this is not the best way to do it or if there is any way it can be improved and I will happily revisit. Thank you

### DIFF
--- a/airmozilla/auth/tests/test_views.py
+++ b/airmozilla/auth/tests/test_views.py
@@ -11,6 +11,7 @@ from nose.tools import ok_, eq_
 
 from airmozilla.auth.browserid_mock import mock_browserid
 from airmozilla.base import mozillians
+from airmozilla.base.tests.testbase import Response
 from airmozilla.main.models import UserProfile
 
 
@@ -85,12 +86,6 @@ NOT_VOUCHED_FOR = """
   ]
 }
 """
-
-
-class Response(object):
-    def __init__(self, content=None, status_code=200):
-        self.content = content
-        self.status_code = status_code
 
 
 class TestViews(TestCase):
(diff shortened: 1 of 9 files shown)


peterbe:
Excellent! At the time of writing, the travis tests haven't finished yet. I'll take a closer look tomorrow.

peterbe:
Welcome to the community @a-buck !

I think you can be bold and delete the JsonResponse class and instead, where it used `return JsonResponse(... json.dumps(dict), ...)` just use `return Response(... dict, ...)` instead.

peterbe:
Clearly you're good at refactoring. Once you're done with this PR, would you mind looking at https://bugzilla.mozilla.org/show_bug.cgi?id=1171160 ?

a-buck:
Thanks @peterbe! No problem I'd be very happy to take a look.

peterbe:
Wow! I totally missed the fact that you had updated the PR. GitHub's notification system isn't perfect. 
If you make changes to a PR and want some eyes on it again, I recommend you write a comment like "Hey @peterbe I addressed those you mentioned in your last comment". 

By the way, WELCOME TO THE AIR MOZILLA COMMUNITY. You are now a fully fledged contributor. If you want to name and photo, please join Mozillians.org and then request to join the [air-mozilla-contributors group](https://mozillians.org/en-US/group/air-mozilla-contributors/).

PR #342, opened 2015-06-03
fixed issue with pagination breaking on gallery

Nolski:
I'm still open to suggestions for a better way to do this. It might be possible to somehow store our own index variable within the ng-repeat model but I haven't been able to find out how.


peterbe:
r+ but you need to change the commit message. 
It should be something like `fixes bug 1170628 - fixed issue with pagination breaking on gallery`

PR #347, opened 2015-06-05
Bug 981128 - Also in channel or Related events

blossomica:
@peterbe - This is not functional right now. I'll remove the prints from it later on and we said the html suppose to be on event.html, opposed to it being a new file as I did, but I didn't really do that. I'ld like to have some feedback on this on if it is going towards the right direction or it is totally off. Thank you :)

### DIFF
--- a/airmozilla/main/static/main/js/es_trigger.js
+++ b/airmozilla/main/static/main/js/es_trigger.js
@@ -1,7 +1,0 @@
-$(document).ready(function() {
-
-    $.get('{% url es_to_template %}', function(response) {
-    // Action to be added if necessary
-  });
-      
-});
--- a/airmozilla/main/static/main/js/es_trigger.js
+++ b/airmozilla/main/static/main/js/es_trigger.js
@@ -1,0 +1,7 @@
+$(document).ready(function() {
+
+    $.get('{% url es_to_template %}', function(response) {
+    // Action to be added if necessary
+  });
+      
+});
--- a/airmozilla/main/templates/main/es.html
+++ b/airmozilla/main/templates/main/es.html
@@ -1,5 +1,2 @@
-<div>
 <h3>Related Events</h3>
 {% include "main/_events.html" %}
-</div>
-
--- a/airmozilla/main/templates/main/es.html
+++ b/airmozilla/main/templates/main/es.html
@@ -1,5 +1,5 @@
 <div>
 <h3>Related Events</h3>
-{% include "main/_events.html %}
+{% include "main/_events.html" %}
 </div>
 
--- a/airmozilla/main/templates/main/es.html
+++ b/airmozilla/main/templates/main/es.html
@@ -1,0 +1,5 @@
+<div>
+<h3>Related Events</h3>
+{% include "main/_events.html %}
+</div>
+
--- a/airmozilla/main/views.py
+++ b/airmozilla/main/views.py
@@ -1231,15 +1231,17 @@ def related_content(request, slug):
                       .filter(id__in=ids)
 
     curated_groups_map = collections.defaultdict(list)
+    events_list = list(events)
+    events_sorted = sorted(events_list, key=lambda e: ids.index(e.id))
 
     def get_curated_groups(event):
         return curated_groups_map.get('event_id')
 
     context = {
-        'events': events,
+        'events': events_sorted,
         'get_curated_groups': get_curated_groups,
     }
-    print context
+
     return render(request, 'main/es.html', context)
 
 
--- a/airmozilla/main/views.py
+++ b/airmozilla/main/views.py
@@ -1203,19 +1203,23 @@ def related_content(request, slug):
         }
     }
 
+    query_ = {
+        'bool': {
+            'should': [mlt_query1, mlt_query2],
+        }
+    }
+
     if request.user.is_active:
         if is_contributor(request.user):
             query = {
                 'fields': fields,
-                'query': {
-                    'bool': {
-                        'should': [mlt_query1, mlt_query2],
-                    }
-                },
+                'query': query_,
                 'filter': {
-                    'must_not': {
-                        'term': {
-                            'privacy': Event.PRIVACY_COMPANY
+                    'bool': {
+                        'must_not': {
+                            'term': {
+                                'privacy': Event.PRIVACY_COMPANY
+                            }
                         }
                     }
                 }
@@ -1223,20 +1227,12 @@ def related_content(request, slug):
         else:
             query = {
                 'fields': fields,
-                'query': {
-                    'bool': {
-                        'should': [mlt_query1, mlt_query2],
-                    }
-                }
+                'query': query_
             }
     else:
         query = {
             'fields': fields,
-            'query': {
-                'bool': {
-                    'should': [mlt_query1, mlt_query2],
-                }
-            },
+            'query': query_,
             "filter": {
                 "bool": {
                     "must": {
@@ -1246,15 +1242,9 @@ def related_content(request, slug):
             }
         }
 
-    query['from'] = 0
-    query['size'] = settings.RELATED_CONTENT_SIZE
     ids = []
     hits = es.search(query, index=index)['hits']
     for doc in hits['hits']:
-        # print "DOC"
-        # from pprint import pprint
-        # pprint(doc)
-        # print "\t", repr(doc['_source']['title']), doc['_id'], doc['_score']
         ids.append(int(doc['_id']))
 
     events = Event.objects.scheduled_or_processing() \
--- a/airmozilla/main/views.py
+++ b/airmozilla/main/views.py
@@ -1242,7 +1242,8 @@ def related_content(request, slug):
         'get_curated_groups': get_curated_groups,
     }
 
-    return render(request, 'main/es.html', context)
+    return_str = render(request, 'main/es.html', context)
+    return http.HttpResponse(return_str)
 
 
 def channels(request):
--- a/airmozilla/main/views.py
+++ b/airmozilla/main/views.py
@@ -1163,26 +1163,26 @@ def related_content(request, slug):
     event = get_object_or_404(Event, slug=slug)
 
     es = pyelasticsearch.ElasticSearch(settings.RELATED_CONTENT_URL)
-    es.refresh()
+    mlt_query = {
+        "more_like_this": {
+            "fields": ["tags", "title", "channels"],
+            "docs": [
+                {
+                    "_index": "events",
+                    "_type": "event",
+                    "_id": event.id
+                }],
+            "min_term_freq": 1,
+            "max_query_terms": 5,
+        }
+    }
 
     if request.user.is_active:
         if is_contributor(request.user):
             query = {
                 "query": {
                     "filtered": {
-                        "query": {
-                            "more_like_this": {
-                                "fields": ["tags", "title", "channels"],
-                                "docs": [
-                                    {
-                                        "_index": "events",
-                                        "_type": "event",
-                                        "_id": event.id
-                                    }],
-                                "min_term_freq": 1,
-                                "max_query_terms": 5,
-                            }
-                        },
+                        "query": mlt_query,
                         "filter": {
                             "must_not": {
                                 "term": {
@@ -1195,37 +1195,13 @@ def related_content(request, slug):
             }
         else:
             query = {
-                "query": {
-                    "more_like_this": {
-                        "fields": ["tags", "title", "channels"],
-                        "docs": [
-                            {
-                                "_index": "events",
-                                "_type": "event",
-                                "_id": event.id
-                            }],
-                        "min_term_freq": 1,
-                        "max_query_terms": 5,
-                    }
-                }
+                "query": mlt_query
             }
     else:
         query = {
             "query": {
                 "filtered": {
-                    "query": {
-                        "more_like_this": {
-                            "fields": ["tags", "title", "channels"],
-                            "docs": [
-                                {
-                                    "_index": "events",
-                                    "_type": "event",
-                                    "_id": event.id
-                                }],
-                            "min_term_freq": 1,
-                            "max_query_terms": 5,
-                        }
-                    },
+                    "query": mlt_query,
                     "filter": {
                         "bool": {
                             "must": {
@@ -1242,18 +1218,17 @@ def related_content(request, slug):
 
     for doc in hits['hits']:
         print "\t", repr(doc['_source']['title']), doc['_id']
-        ids.append(doc['_id'])
-    ids2 = [int(x) for x in ids]
+        ids.append(int(doc['_id']))
 
     if request.user.is_active:
         if is_contributor(request.user):
             events = Event.objects \
-                          .filter(id__in=ids2)
+                          .filter(id__in=ids)
         else:
             events = Event.objects.scheduled_or_processing()
     else:
         events = Event.objects \
-                      .filter(id__in=ids2)
+                      .filter(id__in=ids)
 
     curated_groups_map = collections.defaultdict(list)
 
--- a/airmozilla/main/views.py
+++ b/airmozilla/main/views.py
@@ -1230,8 +1230,7 @@ def related_content(request, slug):
         ids.append(int(doc['_id']))
 
     events = Event.objects.scheduled_or_processing() \
-        .filter(id__in=ids) \
-        .exclude(status=Event.STATUS_REMOVED)
+        .filter(id__in=ids)
 
     if request.user.is_active:
         if is_contributor(request.user):
--- a/airmozilla/main/views.py
+++ b/airmozilla/main/views.py
@@ -1165,7 +1165,7 @@ def related_content(request, slug):
     es = pyelasticsearch.ElasticSearch(settings.RELATED_CONTENT_URL)
     mlt_query = {
         "more_like_this": {
-            "fields": ["tags", "title", "channels"],
+            "fields": ["title", "tags", "channels"],
             "docs": [
                 {
                     "_index": "events",
@@ -1173,7 +1173,7 @@ def related_content(request, slug):
                     "_id": event.id
                 }],
             "min_term_freq": 1,
-            "max_query_terms": 5,
+            "max_query_terms": 20,
         }
     }
 
@@ -1213,6 +1213,8 @@ def related_content(request, slug):
             }
         }
 
+    query['from'] = 0
+    query['size'] = settings.RELATED_CONTENT_SIZE
     ids = []
     hits = es.search(query, index='events')['hits']
 
@@ -1220,15 +1222,12 @@ def related_content(request, slug):
         print "\t", repr(doc['_source']['title']), doc['_id']
         ids.append(int(doc['_id']))
 
+    events = Event.objects.scheduled_or_processing().filter(id__in=ids)
     if request.user.is_active:
         if is_contributor(request.user):
-            events = Event.objects \
-                          .filter(id__in=ids)
-        else:
-            events = Event.objects.scheduled_or_processing()
+            events = events.exclude(privacy=Event.PRIVACY_COMPANY)
     else:
-        events = Event.objects \
-                      .filter(id__in=ids)
+        events = events.filter(privacy=Event.PRIVACY_PUBLIC)
 
     curated_groups_map = collections.defaultdict(list)
     events = sorted(events, key=lambda e: ids.index(e.id))
--- a/airmozilla/main/views.py
+++ b/airmozilla/main/views.py
@@ -59,8 +59,6 @@ from airmozilla.staticpages.views import staticpage
 from . import cloud
 from . import forms
 
-import pyelasticsearch
-
 
 def debugger__(request):  # pragma: no cover
     r = http.HttpResponse()
@@ -1206,23 +1204,23 @@ def related_content(request, slug):
     }
 
     if request.user.is_active:
-       if is_contributor(request.user):
+        if is_contributor(request.user):
             query = {
-            'fields': fields,
-            'query': {
-                'bool': {
-                    'should': [mlt_query1, mlt_query2],
-                }
-            },
-            "filter": {
-                "must_not": {
-                    "term": {
-                        "privacy": Event.PRIVACY_COMPANY
+                'fields': fields,
+                'query': {
+                    'bool': {
+                        'should': [mlt_query1, mlt_query2],
+                    }
+                },
+                'filter': {
+                    'must_not': {
+                        'term': {
+                            'privacy': Event.PRIVACY_COMPANY
+                        }
                     }
                 }
             }
-        }
-       else:
+        else:
             query = {
                 'fields': fields,
                 'query': {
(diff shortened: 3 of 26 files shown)


airmozilla/main/templates/main/es.html
```diff
@@ -0,0 +1,25 @@
+<!-- This is a kind of draft mode and probably has way too much info than needed, is this suppose to be at event.html? -->
+<div>
+<h2 class="section-title">{{ _('Related Events') }}</h2>
+      {% else %}
```
peterbe: where's the `if`? :)

airmozilla/main/templates/main/es.html
```diff
@@ -0,0 +1,25 @@
...
+          {% endif %}
+            <h2 class="section-title">{{ channel.name }}</h2>
+        {% endif %}
+        {% if channel.description %}
+        <div class="channel-description">
+        {{ channel.description|safe }}
+        </div>
+        {% endif %}
+</div>
+
```
peterbe: You get. This is how this template is supposed to work. No `extends` or `blocks`. 
However, I think you don't need to worry about channels at all. This template could basically just be this:
```html
<h3>Related Events</h3>
{% include "main/_events.html" %}
```

airmozilla/main/views.py
```diff
@@ -482,6 +485,45 @@ def get(self, request, slug):
...
+        es.index(
+            'events',
+            'event',
+            {
+                'title': event.title,
+                'tags': [x.name for x in event.tags.all()],
+                'channels': [x.name for x in event.channels.all()],
+            },
+            id=event.id,
+        )
```
peterbe: This for loop doesn't belong here at all. But I guess you knew that.

airmozilla/main/views.py
```diff
@@ -482,6 +485,45 @@ def get(self, request, slug):
...
+    print '-' * 100
+    hits = es.search('title: firefox', index='events')['hits']
+    for doc in hits['hits']:
+        event = Event.objects.get(id=doc['_id'])
+        print "\t", repr(event.title), doc['_score']
+   
+    template_name = 'main/es.html'
+    ids = [int(x) for x in id.split(',')]
+    events = Event.objects.filter(id__in=ids)
+    return render(request, template_name, events)
```
peterbe: Independent of what goes above this line, this line needs to be:
```python
return render(request, 'main/es.html', context)
```
And the line above should be:
```python
context = {
    'events': events,
}
```
If you do that, you're going to make big progress here.

airmozilla/main/views.py
```diff
@@ -482,6 +485,45 @@ def get(self, request, slug):
...
+
+    print '-' * 100
+    hits = es.search('title: firefox', index='events')['hits']
+    for doc in hits['hits']:
+        event = Event.objects.get(id=doc['_id'])
+        print "\t", repr(event.title), doc['_score']
+   
+    template_name = 'main/es.html'
+    ids = [int(x) for x in id.split(',')]
+    events = Event.objects.filter(id__in=ids)
```
peterbe: So, the order might be something that's driven from ES. Basically, once you've figured out how to tell ES to score the results and have them sorted by that, you'll want to preserve the order that ES returned them in. One way to do that would be like this:
```python
hits = es.search('title: firefox', index='events')['hits']
ids = []
for doc in hits['hits']:
    ids.append(doc['_id'])
events = Event.objects.filter(id__in=ids)
events.sort(lambda e: ids.index(e.id))
```
That way, you get them in the order that ES had them. And you're only doing 1 django ORM query. 

That, would get you going. There's going to be a lot of necessary optimization but that's something that can wait.

blossomica:
@peterbe - I didn't understand much about the "for loop" part. I did some edits there from what I thought was the edit but if it needs to totally go you can let me know.

peterbe:
Which "for loop" part? You mean where your code indexes things? 
Basically, your view currently does some WRITEs and some READs. It should only do READs. The WRITEs belong in some other piece of code. E.g. cron job.

blossomica:
@peterbe - Thanks for the explanation. I fixed that.

blossomica:
@peterbe - I did some changes here but I see one of the errors in the error log is that there is no module called pyelasticsearch along with some other test errors on things I didn't really change.

peterbe:
You need to add pyelasticsearch to the `requirements.txt` file. 
I recommend you use [peepin](https://pypi.python.org/pypi/peepin). Like this:
```
pip install peep peepin
peepin pyelasticsearch
git diff
```

peterbe:
So you now have (apart from the `requirements.txt` thing) a really promising start. 
I think we should do the following

1. Your view function is not a view function. It's a method on a class. It needs to be its own function and you need to hook it up in `main/urls.py`.

2. You need to start working on the jQuery code that calls this view function. E.g. something that needs to happen on loading and event. Perhaps look at download.js for inspiration. 

3. Add a setting in `settings/base.py` that is something like `USE_RELATED_CONTENT = False` which you can override. Then, we'll be able to use this so that, by default (in prod) it doesn't run any of your code yet. This way we'll be able to possible land your code even though it's not entirely finished. 

4. You need to start working on a cron job function that indexes the events. Perhaps look at the `cron.py` file(s) for how it's done with autocompeter_update and autocompeter_reset.

peterbe:
Let's get rid of the "es" name. E.g Instead of "es_to_template" call it what it is "related_content".

blossomica:
@peterbe - I added some edits so you can check those whenever you get the chance.

peterbe:
That's really good!! You're totally on the right track. 
Now look into `crontab.tpl` too. 

And get started on indexing the privacy and make sure you use that query accordingly.

blossomica:
@peterbe - I added some more changes. What do you mean by "make sure you use that query accordingly", are you talking of using it in the more-like-this queries?

peterbe:
What I meant about the query is basically this:

When we index events we need to index them under "public", "contributors" or "company". 

When an anonymous user visits the related_content view it should do a fancy more_like_this query against ES but it should ALWAYS require that privacy=public. 

When a Mozillian contributor signs in, it too should do a fancy more_like_this query against ES but it HAS NOT filter on `privacy==public OR privacy==contributors`. 

When a signed in non-contributor signs in and searches, you can ignore the whole filtering on privacy. 

A good example of seeing this logic might just be the home page: https://github.com/mozilla/airmozilla/blob/master/airmozilla/main/views.py#L88-L96

Take a couple of minutes to study that code to get a feel for what it does. That code is all about the Django ORM but you're going to need to do the equivalent with an ES query.

blossomica:
@peterbe - I submitted some fixes for the privacy, you can have a look when you get the chance. Thank you.

peterbe:
No, it's unfortunately not right. And I can understand how you made this easy mistake. The filtering needs to happen on the ES query. The reason is that we rely on ES to do the ordering. 
So, we know we want to display, say, 4 related events. Imagine that ES finds 100 events that "match" in some way but you don't want to turn 100 IDs into event objects so you take the top 10 IDs (from ES's ordering by score) and turn those into real Django Event objects. And then you do a filtering based on privacy. What might happen then is that you only get 3 Event objects. Or 2. Or 0. 

If instead you force in the filtering early to the ES query. If the user can only see Public events, then you'll only get matching IDs that belong to events that are Public. 

Having said all of that, it does also mean we might want to do extra filtering on the Event objects too. **Just to be sure**. I.e. Suppose we send an event into ES as marked as privacy=Public. But then someone decides to change it to privacy=Company. Until the next cron job fixes this, you have a mismatch at your hands. It's privacy=Public in ES and privacy=Company in the postgres database. If that happens, the anonymous user should NOT see it. So basically, keep the code you added to that last commit but remember that that's only for a second extra protection.

airmozilla/manage/related.py
```diff
@@ -0,0 +1,21 @@
+import pyelasticsearch
+
+from airmozilla.main.models import *
+
+
+def indexing(self):
+    es.refresh()
+    es = pyelasticsearch.ElasticSearch(settings.RELATED_CONTENT_URL)
+    for event in events:
```
peterbe: `events` is not defined.

blossomica:
@peterbe : I tried to fix the filtering you said but when i try to run ES it gives me TypeError "related_content() takes exactly 3 arguments (2 given)" , how can I figure out what values are given when I run .../event/related_content/ in the browser? Also any feedback on this newest PR?

peterbe:
All those big dicts of `query:` and `filter:` etc. need to be refactored so nothing that doesn't need to be is repeated.

blossomica:
@peterbe : how does this look like so far? I haven't done the refactoring because it isn't working so far.  I have tried several things but it was giving me invalid syntax so I am still working on that. However I see nothing when I run ES only the title "Related Content" appears.

peterbe:
You're making progress! And you're growing more and more capable now. 

I think we need to change approach a bit. I like that you keep sending me small snippets of your progress but I think you also need to start only sending things in when they are working. Like little milestones instead of little snippets. A lot of changes you should be able to figure out with time and patience by studying the django documentation or studying other apps and solutions within airmozilla to see how they do it. 

You'll need to try to debug why you're not seeing any content. Break things. Take then apart. Lift every stone. Seek deeper. You have time. There is time for learning whilst doing. 
Ping me when you're really stuck and when you do, let's change the format slightly. 

When you get stuck, for example, you can't figure out why you get a certain error. Then take note of how you're debugging that and what things you've tried. Then when you come to me, prefix your stuck'ness situation by explaining what you've tried. 

I think that this way you'll learn more. Instead of me helping you with the solution, I'll help you become better at debugging and testing. Your personal development is more important than Air Mozilla visitors seeing related content :)

This is called "Rubber duck Debugging". It's when you say to a colleague "Hey, can you help me figure out why I keep getting this error? First, let me show you what I do to get the error. Then, let me explain why the code does what it does to get to that strange error."
When you do that, you often find out what's wrong just whilst showing what the problem is. The person you asked for help often just stands there saying nothing. The term comes from the silly example that you have a rubber duck on your computer desk and you pretend it's a person you have to explain the problem to. And by explaining it to someone else you often find new ways of thinking. 

So, time to get down and dirty with django! :)

blossomica:
@peterbe : All right. I'll follow your advice. Thanks.

peterbe:
So, one thing we need to do is change the query of how we use `more_like_this`. Instead of sending `'title': event.title` you need to use `_docs`. See the documentation https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-mlt-query.html

So you'd do something like this:
```
{
    "more_like_this" : {
        "fields" : ["tags", "title", "channels"],
        "docs" : [
        {
            "_index" : "events",
            "_type" : "event",
            "_id" : event.id
        }],
        "min_term_freq" : 1,
        "max_query_terms" : 12
    }
}
```

blossomica:
@peterbe : All right. Thanks for letting me know, I'll look into it :)

blossomica:
@peterbe : This here works haha, opposed to what I posted so far. I see the found events when I run it. I haven't check if it works for "groups" because that's something I haven't checked on yet, but i'll figure out how to do that. I'll like some feedback on this. I get some errors in classes unrelated to what I am doing when I run the code...error from the starred event tests and I don't understand why. I am so glad to be finally able to ask for feedback :)

airmozilla/main/views.py
```diff
@@ -1114,6 +1115,115 @@ def item_pubdate(self, event):
...
+                    "filter": {
+                        "bool": {
+                            "must": {
+                                "term": {"privacy": Event.PRIVACY_PUBLIC}
+                            }
+                        }
+                    }
+                }
+            }
+        }
```
peterbe: See if you can refactor all of this so that it doesn't have to define the `"query"` part each time. The only thing that needs to change is the `"filter"` stuff.

blossomica:
@peterbe : I just did the refactoring.

peterbe:
Cool! So, whats next? Working on the template rendering?

blossomica:
@peterbe : what do you mean by template rendering?

Also I still don't get why the tests for stared events are failing when I have done nothing with them.

Lastly, I see in [ES documentation](https://www.elastic.co/guide/en/elasticsearch/guide/current/sorting.html) that 
>By default, results are returned sorted by relevance—with the most relevant docs first. Later in this chapter, we explain what we mean by relevance and how it is calculated, but let’s start by looking ?at the sort parameter and how to use it.

So I don't really need to do any form of sorting maybe?

peterbe:
> what do you mean by template rendering?

Well, the way it's all going to work is that once you're on an event, it should load a piece of Javascript that makes an AJAX query to get the thumbnails and links of a couple of related events. 

> So I don't really need to do any form of sorting maybe?

Yes. The order of the `ids` list is important. If the value of that list is `[34, 99, 13, 4]` that means that ID=34 is most related and ID=4 least related. When we do `filter(ids__in=[34, 99, 13, 4])` it returns a QuerySet which does NOT have the same order as the list. 
What you need to do is convert the QuerySet to a *list* of events and *that* list you need to order so it matches the order of the IDs. 

The solution lies in studying https://github.com/mozilla/airmozilla/blob/master/airmozilla/starred/views.py#L67

blossomica:
@peterbe : I have submitted some code you could look into. The ajax part is not complete/working, but the other parts function all right.

airmozilla/main/views.py
```diff
@@ -1114,6 +1115,93 @@ def item_pubdate(self, event):
...
+                          .filter(id__in=ids)
+        else:
+            events = Event.objects.scheduled_or_processing()
+    else:
+        events = Event.objects \
+                      .filter(id__in=ids)
+
+    curated_groups_map = collections.defaultdict(list)
+    events_list = list(events)
+    events_sorted = sorted(events_list, key=lambda e: ids.index(e.id))
```
peterbe: You can simplify these two lines with:
```python
events = sorted(events, key=lambda e: ids.index(e.id))
```
(Note you'll need to change line 1197 too.

airmozilla/main/views.py
```diff
@@ -1114,6 +1115,93 @@ def item_pubdate(self, event):
...
+    def get_curated_groups(event):
+        return curated_groups_map.get('event_id')
+
+    context = {
+        'events': events_sorted,
+        'get_curated_groups': get_curated_groups,
+    }
+
+    return_str = render(request, 'main/es.html', context)
+    return http.HttpResponse(return_str)
```
peterbe: These two lines can be just `return render(request, 'main/es.html', context)`

airmozilla/settings/base.py
```diff
@@ -224,6 +224,9 @@ def COMPRESS_JINJA2_GET_ENVIRONMENT():
 # Always use secure cookies
 COOKIES_SECURE = True
 
+#Default for related content
```
peterbe: hash characters for comments should have a space after them. See other comments.

blossomica:
@peterbe - I saw some other comments in base.py without a space after the hash so i'll fix those too.

airmozilla/manage/related.py
```diff
@@ -0,0 +1,30 @@
+import pyelasticsearch
+
+from django.conf import settings
+
+from airmozilla.main.models import Event
+
+
+def indexing():
```
peterbe: "indexing" is not a good word for a function. Function names should be verbs.

blossomica:
@peterbe : I added some fixes for the crontab issue and the other things. Is it okay?

airmozilla/main/views.py
```diff
@@ -1114,6 +1115,90 @@ def item_pubdate(self, event):
         return event.start_time
 
 
+def related_content(request, slug):
+    event = get_object_or_404(Event, slug=slug)
+
+    es = pyelasticsearch.ElasticSearch(settings.RELATED_CONTENT_URL)
+    mlt_query = {
+        "more_like_this": {
+            "fields": ["title", "tags", "channels"],
```
peterbe: Make this a variable. A couple of lines up here in this function. Then, use that to NOT include `channels` if the event you're on is only in the default channel. E.g. something like this:
```python
fields = ['title', 'tags']
if list(event.channels.all()) != [Channel.objects.get(slug=settings.DEFAULT_CHANNEL_SLUG):
    fields.append('channel')
```
Obviously, you might need to format that differently so it fits better.

peterbe: The idea is to not search for events that are on the main channel. It doesn't add any value to the search.

airmozilla/manage/cron.py
```diff
@@ -99,3 +100,15 @@ def autocompeter_update():
         # this number is supposed to match that of the cronjob itself
         since=datetime.timedelta(minutes=10)
     )
+
+
+@cronjobs.register
+# @capture
+def related_content_update():
```
peterbe: "update" is not a good name when its job is to reindex from scratch.

blossomica:
@peterbe - I added the fixes and It seems the search results might have been improved by not comparing the main tag. I checked what you said and it looks as if there is indeed something strange going on. Does this fixes it? If not, I was thinking if I need to boost certain scores or if I need to change the ES query filtering I did altogether. I'd like to know your thoughts on that. Thanks.

peterbe:
So, I'm no ES expert but I think that what you need to do is to instruct ES somehow which **analyzer** to use for various fields. For example, for the `title` field we want full English analyzer (the default) which includes case-**in**sensitivity and stemming ("rocks" == "rocking"). For tags, we don't want that. We want to treat tags as "keywords" I believe. Same for channels. 

For example, if you have a tag called "firefox on android", I think it's going to match an even whose title is "Testing Firefox with an Android phone". It shouldn't do that. Tags should be treated as non-English. Like "xyz123" or "SOME_ERROR_CODE_CONSTANT".

blossomica:
@peterbe: Interesting I didn't know ES has such distinction. I'll look further into that and see what I can come up with. Thank you.

peterbe:
I'm sorry to say this, but; I think you have to read a lot of documentation and study examples on the internet. 

When it comes to setting up analyzers I think you can have a choice. Either you define it first. Before you do any `es.index(...)` or you do it **whilst** you do the `es.index(...)` call.

airmozilla/main/views.py
```diff
@@ -1114,6 +1115,96 @@ def item_pubdate(self, event):
...
+    query['from'] = 0
+    query['size'] = settings.RELATED_CONTENT_SIZE
+    ids = []
+    hits = es.search(query, index='events')['hits']
+
+    for doc in hits['hits']:
+        print "\t", repr(doc['_source']['title']), doc['_id']
+        ids.append(int(doc['_id']))
+
+    events = Event.objects.scheduled_or_processing().filter(id__in=ids)
```
peterbe: Add `.exclude(status=Event.STATUS_REMOVED)`

airmozilla/manage/related.py
```diff
@@ -0,0 +1,48 @@
...
+from django.utils import timezone
+
+from airmozilla.main.models import Event
+
+
+def index(all=False, flush_first=False, since=datetime.timedelta(minutes=10)):
+    es = pyelasticsearch.ElasticSearch(settings.RELATED_CONTENT_URL)
+
+    if flush_first:
+        es.flush()
```
peterbe: This needs to change. We need to flush only the relevant index.

peterbe:
For the testing...

So, we need to make the name of the index used a setting. 
Add to `settings/base.py` something like 
```
ELASTICSEARCH_PREFIX = 'airmozilla'
ELASTICSEARCH_INDEX = 'events'
```

And then in `settings_test.py` we must override this to be:
```
ELASTICSEARCH_INDEX = 'test-events'
```

And in `related.py` you need to use these values now. 
Instead of `es.index('events', ...` it needs to be `es.index(settings.ELASTICSEARCH_INDEX, ...)`

peterbe:
In fact, we need to go deeper. We need to use the prefix too. 
Instead of `es.index(settings.ELASTICSEARCH_INDEX, ...)`
use `es.index(settings.ELASTICSEARCH_PREFIX + settings.ELASTICSEARCH_INDEX, ...)`

That means we potentially have two indexes for the whole of Air Mozilla. 
* `airmozillaevents`
* `airmozillatest-events`

We also need to replace `es.refresh()` with `es.refresh(settings.ELASTICSEARCH_PREFIX + settings.ELASTICSEARCH_INDEX)`
And the same for `es.flush()`

blossomica:
@peterbe : I added the edits suggested here, so you could work with it and check up on the things needed to be looked at. I need to get that working properly as it is buggy but this is the current update.

peterbe:
As you might have seen, I sent you a PR that solves the problem with flushing. 

So, I'm confused. I thought you had a fix that tried to set some analyzers. Did you retract that?
I now get almost "random" results. Definitely not setting the one I clicked *from* in the related content, if you see what I mean.

blossomica:
@peterbe : Yes I retracted that. I couldn't see it doing anything. I couldn't see a difference in the results  so I removed it. I can add it so you can check it too. I was trying to do something else and removed it then I thought it was working...most of the results seemed good till I signed off airmozilla to try as a guest and I started getting the almost random results as well.

airmozilla/main/views.py
```diff
@@ -1114,6 +1115,100 @@ def item_pubdate(self, event):
...
+    ids = []
+    hits = es.search(query, index='events')['hits']
+
+    for doc in hits['hits']:
+        print "\t", repr(doc['_source']['title']), doc['_id'], doc['_score']
+        ids.append(int(doc['_id']))
+
+    events = Event.objects.scheduled_or_processing() \
+        .filter(id__in=ids) \
+        .exclude(status=Event.STATUS_REMOVED)
```
peterbe: Perhaps this is not necessary. After all, the `scheduled_or_processing()` should already do the necessary filtering on status.

airmozilla/main/static/main/js/es_trigger.js
```diff
@@ -0,0 +1,7 @@
+$(document).ready(function() {
+
+    $.get('{% url es_to_template %}', function(response) {
+    // Action to be added if necessary
+  });
+      
+});
```
peterbe: I think you can delete this file now. No?

airmozilla/manage/related.py
```diff
@@ -0,0 +1,44 @@
...
+            'event',
+            {
+                'title': event.title,
+                'privacy': event.privacy,
+                'tags': [x.name for x in event.tags.all()],
+                'channels': [x.name for x in event.channels.all()],
+            },
+            id=event.id,
+
+        )
```
peterbe: So I think it's time to start thinking about doing a bulk index instead. It's going to make little difference if there's only 1-5 events that needs to be indexed but it's going to make a HUGE difference when we ALL. 

What we need to do is make two bulk updates. 
Something like this:
```python
if all:
    events = Event.objects.scheduled_or_processing()
    removed = Event.objects.filter(status=Event.STATUS_REMOVED)
else:
    now = timezone.now()
    events = Event.objects.scheduled_or_processing() \
            .filter(modified__gte=now-since)
    removed = Event.objects.filter(status=Event.STATUS_REMOVED)\
             .filter(modified__gte=now-since)

events_as_list = [
    {
                'id': event.id,
                'title': event.title,
                'privacy': event.privacy,
                'tags': [x.name for x in event.tags.all()],
                'channels': [x.name for x in event.channels.all()],
            }
         for event in events
]
es.bulk(
    (es.index_op(doc, id=doc.pop('id')) for doc in events_as_list),
    index=settings.ELASTICSEARCH_PREFIX + settings.ELASTICSEARCH_INDEX,
    doc_type='event'
)
for event in removed:
    es.delete_op(id=event.id)
```

airmozilla/settings/settings_test.py
```diff
@@ -0,0 +1 @@
+ELASTICSEARCH_INDEX = 'test-events'
```
peterbe: Remember, this line needs to be moved. As discussed.

PR #351, opened 2015-06-08
fixes bug 1172569 - Simplified picture gallery float solution

Nolski:
The diff here is a bit ugly since removing the surrounding `.row` divs made me have to unindent all of the HTML inside of that but the main change is the addition of `'clearfix': $index % 4 == 0` to the ng-class and removal of the old and janky code for creating rows around every 4 elements. This should work much better.


peterbe:
r+ 

I'll merge as soon as Travis is happy.

PR #366, opened 2015-06-14
fixes bug 870015 - background reloading on event state change

a-buck:
Hi @peterbe 

This is a pull request for the [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=870015) you suggested I look at previously such that if the event status for a live event changes to pending or from pending to archived it will reload the page. Thanks for the suggestion it certainly felt a little more challenging!

I look forward to hearing any feedback as I wasn't 100% sure on the best way to do this one - particularly with respect to how the javascript accesses the data it needs from the html page.

I would also like to add a test for the view but I have held off doing that for now as I am still figuring out how to do that.

thanks, Alex


airmozilla/main/static/main/js/eventstatus.js
```diff
@@ -0,0 +1,19 @@
+(function () {
```
peterbe: Use the jQuery onload event handler to be safe. That way it doesn't execute until the `document.onload` really has fired. (as opposed to loading when the javascript file has been loaded). 
The way things work today in airmo, this is basically the same thing but it's safer to be safe. 

So change this line to `$(function() {` and change the very last line to `});`

airmozilla/main/urls.py
```diff
@@ -75,6 +75,9 @@
     url(r'^(?P<slug>[-\w]+)/(?P<id>\d+)/difference/$',
         views.EventRevisionView.as_view(difference=True),
         name='event_difference'),
+    url(r'^eventstatus/(?P<id>\d+)/$',
+        views.event_status,
+        name='event_status'),
```
peterbe: Let's use the slug instead. So make the regex 
```python
r'^(?P<slug>[-\w]+)/status/$'
```
instead. That way, you never have to define the URL somewhere in the HTML for the JS to pick up. You can always find the URL by simply taking `document.location.pathname + 'status/'`, right?

airmozilla/main/static/main/js/eventstatus.js
```diff
@@ -0,0 +1,19 @@
...
+    }
+
+    function refreshIfStatusChanges() {
+        var url = '/eventstatus/' + eventId;
+        $.get(url, function (data) {
+            if (data.event_status != currentEventStatus) {
+                location.reload();
+            }
+        });
+    }
```
peterbe: You're definitely on the right track and I know the bug didn't mention this but I have another idea. 
Basically, why not apply this business logic. 
```
var url = location.pathname + '/status/';
var eventStatus = null;
var interval = setInterval(function() {
    $.get(url).
    .then(function(data) {
        if (eventStatus !== null && eventStatus !== data.status) {
            location.reload();
        }
        eventStatus = data.data;
    })
    .fail(function() {
        console.error('Failed to fetch event status. Stops looking.');
        clearInterval(interval);
    });
}, 10 * 1000);
```
That's untested and rushed together code but the idea is that after 10 seconds it makes the first load and stores what the current status is and does nothing else. 
Then 10 seconds later it checks what the current status is and whatever is, if it's different from the **first status** that's reason enough to reload. 

This way the code works for all sorts of transitions. From being `scheduled` to `processing` and from `processing` to `scheduled`. Or from `scheduled` to `removed`. And your code won't need to get itself involved at all with the names of statuses. It just looks at differences/changes. Any differences/changes.

airmozilla/main/views.py
```diff
@@ -1560,6 +1560,15 @@ def event_livehits(request, id):
     return {'hits': total_hits}
 
 
+@json_view
+def event_status(request, id):
+    try:
+        event = Event.objects.get(pk=id)
+        return {'event_status': event.status}
+    except Event.DoesNotExist:
+        return {'event_status': ''}
```
peterbe: Let errors be errors :)
I would just let it be a plain 404 Not Found if someone supplied the wrong parameter. 
E.g.
```python
def event_status(request, id):
    return {'event_status': get_object_or_404(Event, id=id).status}
```

a-buck:
Great feedback @peterbe!

So I’ve changed the api from /eventstatus to /status.
I’ve changed it to use the slug instead which seems pretty nice.

You mentioned in your comment on eventstatus.js that after 10 seconds it should store the current status. Why wait 10 seconds to store the current status? The local event status is not going to change so why not load it up front? I think it makes for slightly cleaner code as the if statement doesn't need a null check.

After finishing going back and forth pull requests like this? Is it preferred that commits and squashed or that they are left as is?

Cheers!

peterbe:
The reason for waiting a total of 20 seconds before the first chance of ever reloading was to keep the code simple. That way, the template doesn't need to pack any extra information. It's less noise and less wondering by developers "what's that for?!" The little `eventstatus.js` plus the new JSON endpoint is all it would need. 
Lastly, in your code you depend on `$('.islive')` which isn't a great container for the purpose of the `data-status`. If it was `$('.eventinformation')` it'd make more sense to see an attribute on it called `data-status`. But most importantly, this ties us to one particular state. The "live" state. Imagine a laptop without a screensaver that first clicked on an Upcoming event, then the owner of the laptop disappeared for a long lunch. During the course of the laptop owners disappearance (or perhaps he's lying in bed away from the keyboard) the event will eventually go from Upcoming to Live and then from Live to Archived. First no video, then a live stream, then an archived html5 video. 
By *not* being tied to "live", it should work if all we care about is if the status changed from one to another. 

> After finishing going back and forth pull requests like this, is it preferred that commits and squashed or that they are left as is?

Well, the practice I stick to is that you wait for a formal comment that says r+. When you get that, you can rebase and squash down to 1 (or more if there are different directions within the PR) with a nice git commit message. That'll trigger one last extra Travis build (now being based on the latest master) and then I'll merge it.

airmozilla/main/static/main/js/eventstatus.js
```diff
@@ -0,0 +1,22 @@
+$(function () {
+    var TEN_SECONDS = 10000;
```
peterbe: This is a unfortunate variable name. It makes it "impossible" to change our minds to 12 seconds some day as we'd also need to change from `TEN_SECONDS` to `TWELVE_SECONDS` so what was the point of making it a variable?

Also, a variable that is only used once is usually an anti-pattern. The purpose of variables is re-use so you never type the value more than once. 

Shorter code with less stuff in it is easier to read. This file is only 22 lines long but imagine you're eyes are on the `setInterval(function() {... }, SOME_LENGTH)` line and you're debugging. Then you have to move your eyes (or use your editors Find function) to jump away from what you were doing to find out the value of `SOME_LENGTH`. It's much better to keep the code "compact". I.e. lines that are relevant to each other (and only each other) close. 

What I would do in this case is to just keep it simple with:
```javascript
var interval = setInterval(function () {
    refreshIfStatusChanges();
}, 1000000);
```

peterbe: Ha! I just made my next point for myself. I accidentally typo'ed 100000. One 0 too many. It's easier to read `10 * 1000` and it's a common convention when specifying milliseconds. Already calculating that `10 * 1000` equals `10000` is not a good optimization.

airmozilla/main/static/main/js/eventstatus.js
```diff
@@ -0,0 +1,22 @@
...
+                if (currentEventStatus !== data.status) {
+                    location.reload();
+                }
+            })
+            .fail(function () {
+                console.error("Failed to fetch event status. Stops looking.");
+                clearInterval(interval);
+            });
+    }
+});
```
peterbe: You're definitely on the right track but consider refactoring here a bit. Everything that is only ever used once, don't bother defining it first. 
For the example, the `url`. And the `refreshIfStatusChanges` function.

airmozilla/main/views.py
```diff
@@ -1560,6 +1560,11 @@ def event_livehits(request, id):
     return {'hits': total_hits}
 
 
+@json_view
+def event_status(request, slug):
+    return {'status': get_object_or_404(Event, slug=slug).status}
```
peterbe: Excellent!

peterbe:
You now need to also write a test for that new view function (`event_status`). **You** should be really good at finding a the right file for adding such a new test :)

You only need to write 1 test. One where you get a 404 and one where you get some JSON.

a-buck:
Hey @peterbe, I have made the changes you suggested. 

-refactored `eventstatus.js` and made it depend on `$(‘content-entry’)` for the event-status instead of depending on `$(‘islive’)`
-added a test for the `event_status` view that tests both getting content back, and that there is a 404 if an invalid resource is requested.

Really appreciate all the feedback -  I’m learning a lot :)

peterbe:
Excellent! r+

Now please do the rebase and squash thing and I'll merge it.

a-buck:
Great! @peterbe those commits are now squashed.

PR #375, opened 2015-06-23
Popcorn API

Nolski:


Nolski:
This is my first shot at creating a Django API in a while so feel free to tear it apart and let me know if anything should be changed.

airmozilla/popcorn/tests/test_views.py
```diff
@@ -0,0 +1,51 @@
...
+        event = Event.objects.get(title=self.title)
+        self._attach_file(event, self.main_image)
+
+    @mock.patch('requests.head')
+    def test_meta_data_api(self, rhead):
+
+        def mocked_head(url, **options):
+            return Response(
+                '',
+                200,
```
peterbe: That should be 302

airmozilla/popcorn/tests/test_views.py
```diff
@@ -0,0 +1,51 @@
...
+        rhead.side_effect = mocked_head
+
+        event = Event.objects.get(title='Test event')
+        url = reverse('popcorn:event_meta_data')
+
+        response = self.client.get(url, {'slug': event.slug})
+        content = json.loads(response.content)
+
+        eq_(response.status_code, 200)
+        eq_(content['status'], 'scheduled')
```
peterbe: Instead of writing the string, use `Event.STATUS_SCHEDULED` here

airmozilla/popcorn/tests/test_views.py
```diff
@@ -0,0 +1,51 @@
...
+        ok_('png' in content['preview_img'])
```
peterbe: Just do `ok_(content['preview_img'])`

airmozilla/popcorn/tests/test_views.py
```diff
@@ -0,0 +1,51 @@
...
+from nose.tools import eq_, ok_
+
+from funfactory.urlresolvers import reverse
+
+
+class TestPopcornEvent(DjangoTestCase):
+    fixtures = ['airmozilla/manage/tests/main_testdata.json']
+    main_image = 'airmozilla/manage/tests/firefox.png'
+    title = 'Test event'
+    location = 'http://localhost'
```
peterbe: Since it's only ever used in one place, move these two things into the test. 
Leave the `main_image` since its value actually matters.

airmozilla/popcorn/tests/test_views.py
```diff
@@ -0,0 +1,51 @@
...
+        url = reverse('popcorn:event_meta_data')
+
+        response = self.client.get(url, {'slug': event.slug})
+        content = json.loads(response.content)
+
+        eq_(response.status_code, 200)
+        eq_(content['status'], 'scheduled')
+        ok_('png' in content['preview_img'])
+        eq_(content['description'], 'sadfasdf')
+        eq_(content['title'], self.title)
```
peterbe: instead of `self.title` it should be `event.title`

airmozilla/popcorn/views.py
```diff
@@ -0,0 +1,48 @@
+import requests
+
+from django.shortcuts import get_object_or_404
+
+from jsonview.decorators import json_view
```
peterbe: `jsonview` and `requests` are non-django third-party libs so they belong together grouped as import statements.

airmozilla/popcorn/views.py
```diff
@@ -0,0 +1,48 @@
...
+
+        video_url = 'https://vid.ly/{0}?content=video&format={1}'.format(
+            tag, video_format
+        )
+
+    headers = {
+        'User-Agent': request.META.get('HTTP_USER_AGENT')
+    }
+
+    r = requests.head(video_url, headers=headers)
```
peterbe: If the above if statement isn't true then `video_url` is None and you could do `requests.head(None, headers=headers)` which would crash.

airmozilla/popcorn/views.py
```diff
@@ -0,0 +1,48 @@
...
+
+    return {
+        'title': event.title,
+        'status': event.status,
```
peterbe: Why do you need this?

airmozilla/popcorn/views.py
```diff
@@ -0,0 +1,48 @@
...
+        'description': event.description,
```
peterbe: Every event has a `short_description` and a `description`. The logic is, when you browse events, that we display the `short_description` and if it's empty we show the first couple of words (e.g. 25 words) from the `description`.

peterbe: Let's just change it to `'description': event.short_description or event.description,`

airmozilla/popcorn/views.py
peterbe: `r` is a bad variable name.

peterbe: We'd need to also check that we get the 302 error code too. If you get a 404 we need to fail in a better way.

peterbe:
You need to add a `response.set_header('Cross....`

airmozilla/popcorn/views.py
Nolski: Why would we get a 404 for a vidly video if we are showing that vidly video on our site? Should we throw a 404 if vidly is messed up somehow?

airmozilla/popcorn/views.py
Nolski: I can't think of a use case for popcorn editor so I'll remove this for now.

airmozilla/popcorn/views.py
Nolski: I think if we want to wrap this for safety is to throw a 503 if anything happens making that second request to vidly.

Nolski:
Updated with changes to response

peterbe:
If travis passes, r+
Time to rebase and squash. But first you need to attach this to a bug. 
Feel free to assign https://bugzilla.mozilla.org/show_bug.cgi?id=1178377 to yourself. 
THen make your commit message something like `fixes bug 1178377 - bla bla`

Nolski:
assigned, rebased and squashed.

PR #395, opened 2015-07-13
Editor load video

Nolski:
This updates the video editor page to now auto load the video into the editor.

This PR includes changes to popcorn-editor.js as well. To mess around the UI changes head to http://{domain}/popcorn/edit/{slug}

### DIFF
--- a/airmozilla/popcorn/static/popcorn/PopcornEditor/src/PopcornEditor.js
+++ b/airmozilla/popcorn/static/popcorn/PopcornEditor/src/PopcornEditor.js
@@ -19,7 +19,7 @@ var PopcornEditor = (function () {
   };
 
   // List of events that PopcornEditor supports
-  PopcornEditor.events = {save: 'save'};
+  PopcornEditor.events = {save: 'save', loaded: 'loaded'};
 
   /**
    * Sets the given handler as the handler for the event
@@ -42,6 +42,85 @@ var PopcornEditor = (function () {
          data: data,
          type: 'load'
       }, window.location.origin);
+  };
+
+  /**
+   * Given a javascript object which fits the schema defined below, popcorn
+   * editor will load that video into the editor.
+   *
+   * @param video : javascript object of video
+   */
+  PopcornEditor.createTemplate = function (video) {
+    var videoUrl = video.url;
+    data = {
+        "template": "basic",
+        "background": "#FFFFFF",
+        "data": {
+            "targets": [{
+                "id": "Target0",
+                "name": "video-container",
+                "element": "video-container",
+            }],
+            "media": [{
+                "id": "Media0",
+                "name": "Media0",
+                "url": "#t=,30",
+                "target": "video",
+                "duration": video.duration,
+                "popcornOptions": {
+                    "frameAnimation": true,
+                },
+                "controls": true,
+                "tracks": [{
+                    "name": "",
+                    "id": "0",
+                    "order": 0,
+                    "trackEvents": [{
+                        "id": "TrackEvent0",
+                        "type": "sequencer",
+                        "popcornOptions": {
+                            "start": 0,
+                            "source": [video.url],
+                            "fallback": "",
+                            "denied": false,
+                            "end": video.duration,
+                            "from": 0,
+                            "title": video.title,
+                            "type": "AirMozilla",
+                            "thumbnailSrc": video.thumbnail,
+                            "duration": video.duration,
+                            "linkback": "",
+                            "contentType": "",
+                            "hidden": false,
+                            "target": "video-container",
+                            "left": 0,
+                            "top": 0,
+                            "width": 100,
+                            "height": 100,
+                            "volume": 100,
+                            "mute": false,
+                            "zindex": 1000,
+                            "id": "TrackEvent0"
+                        },
+                        "track": "0",
+                        "name": "TrackEvent0"
+                    }]
+                }],
+                "clipData": {
+                    videoUrl : {
+                        "type": video.type,
+                        "title": video.title,
+                        "source": video.url,
+                        "thumbnail": video.thumbnail,
+                        "duration": video.duration
+                    }
+                },
+                "currentTime": 0,
+            }]
+        },
+        "tags": ["popcorn"],
+    }
+    return data;
   };
 
   window.addEventListener('message', function (e) {
--- a/airmozilla/popcorn/static/popcorn/PopcornEditor/src/editor/media-gallery-editor.js
+++ b/airmozilla/popcorn/static/popcorn/PopcornEditor/src/editor/media-gallery-editor.js
@@ -3,8 +3,10 @@
  * obtain one at https://raw.github.com/mozilla/butter/master/LICENSE */
 
 define( [ "util/lang", "util/uri", "util/xhr", "util/keys", "util/mediatypes", "editor/editor",
- "util/time", "util/dragndrop", "analytics", "l10n!../../{{lang}}/layouts/media-editor.html", "text!../../api/butterconfig" ],
-  function( LangUtils, URI, XHR, KeysUtils, MediaUtils, Editor, Time, DragNDrop, analytics, EDITOR_LAYOUT, CONFIG ) {
+ "util/time", "util/dragndrop", "analytics", "l10n!../../{{lang}}/layouts/media-editor.html",
+ "text!../../api/butterconfig", "events/event" ],
+  function( LangUtils, URI, XHR, KeysUtils, MediaUtils, Editor, Time, DragNDrop,
+           analytics, EDITOR_LAYOUT, CONFIG, Event) {
 
   CONFIG = JSON.parse(CONFIG);
 
@@ -284,11 +286,11 @@ define( [ "util/lang", "util/uri", "util/xhr", "util/keys", "util/mediatypes", "
     resetInput();
   }
 
-  function onSuccess( data ) {
+  function onSuccess( data, imported ) {
     var el = _GALLERYITEM.cloneNode( true ),
         source = data.source;
 
-    if ( !_media.clipData[ source ] ) {
+    if ( !_media.clipData[ source ]  || imported !== undefined) {
       _media.clipData[ source ] = data;
       _butter.dispatch( "mediaclipadded" );
 
@@ -700,6 +702,16 @@ define( [ "util/lang", "util/uri", "util/xhr", "util/keys", "util/mediatypes", "
     _searchInput.addEventListener( "keydown", onEnter );
 
     _addBtn.addEventListener( "click", onAddMediaClick );
+
+    Event.listen('load', function (data) {
+      var clipData = data.data.media[0].clipData,
+          clip;
+
+      for (url in clipData) {
+        clip = clipData[url];
+        onSuccess(clip, true);
+      }
+    });
   }
 
   Editor.register( "media-editor", null, function( rootElement, butter ) {
--- a/airmozilla/popcorn/static/popcorn/js/editor.js
+++ b/airmozilla/popcorn/static/popcorn/js/editor.js
@@ -1,2 +1,20 @@
-// Initialize the editor with the div id and path to Popcorn Editor.
-PopcornEditor.init('editor', '/static/popcorn/PopcornEditor/editor.html');
+$(document).ready(function () {
+  PopcornEditor.listen(PopcornEditor.events.loaded, function () {
+    $.get($('#editor').data('url'), {
+        slug: $('#editor').data('slug')
+    })
+    .done(function (response) {
+      if(response.data) {
+        PopcornEditor.loadInfo(response.data);
+      } else {
+        PopcornEditor.loadInfo(PopcornEditor.createTemplate(response.metadata));
+      }
+    })
+    .fail(function() {
+      console.warn("Unable to load popcorn data :(");
+      console.error.apply(console, arguments);
+    });
+  })
+  // Initialize the editor with the div id and path to Popcorn Editor.
+  PopcornEditor.init('editor', '/static/popcorn/PopcornEditor/editor.html');
+});
--- a/airmozilla/popcorn/templates/popcorn/editor.html
+++ b/airmozilla/popcorn/templates/popcorn/editor.html
@@ -23,7 +23,7 @@ Air Mozilla | Mozilla, in Video
 {% endblock %}
 
 {% block full_width_content %}
-<div id="editor"></div>
+<div id="editor" data-slug="{{ slug }}" data-url="{{ url('popcorn:popcorn_data') }}"></div>
 {% endblock %}
 
 {% block content_main %}{% endblock %}
--- a/airmozilla/popcorn/tests/test_views.py
+++ b/airmozilla/popcorn/tests/test_views.py
@@ -1,5 +1,6 @@
 from airmozilla.base.tests.testbase import DjangoTestCase, Response
 from airmozilla.main.models import Event
+from airmozilla.popcorn.models import PopcornEdit
 
 import mock
 import json
@@ -37,13 +38,15 @@ class TestPopcornEvent(DjangoTestCase):
         rhead.side_effect = mocked_head
 
         event = Event.objects.get(title='Test event')
+        url = reverse('popcorn:event_meta_data')
+
+        response = self.client.get(url, {})
+        eq_(response.status_code, 400)
         event.template.name = 'this is a vid.ly video'
         event.template.save()
 
         event.template_environment = {'tag': 'abc123'}
         event.save()
-
-        url = reverse('popcorn:event_meta_data')
 
         response = self.client.get(url, {'slug': event.slug})
         content = json.loads(response.content)
@@ -54,3 +57,93 @@ class TestPopcornEvent(DjangoTestCase):
         eq_(content['title'], event.title)
         eq_(content['video_url'], location)
         eq_(response['Access-Control-Allow-Origin'], '*')
+
+    @mock.patch('requests.head')
+    def test_popcorn_data(self, rhead):
+        location = 'http://localhost'
+
+        def mocked_head(url, **options):
+            return Response(
+                '',
+                302,
+                headers={
+                    'location': location
+                }
+            )
+
+        rhead.side_effect = mocked_head
+
+        event = Event.objects.get(title='Test event')
+        event.template.name = 'this is a vid.ly video'
+        event.template.save()
+
+        event.template_environment = {'tag': 'abc123'}
+        event.save()
+
+        url = reverse('popcorn:popcorn_data')
+
+        response = self.client.get(url, {'slug': event.slug})
+        # because we're not logged in
+        eq_(response.status_code, 302), rhead
+
+        self._login()
+
+        response = self.client.get(url)
+        # because there is no slug
+        eq_(response.status_code, 400)
+
+        response = self.client.get(url, {'slug': event.slug})
+        eq_(response.status_code, 200)
+
+        content = json.loads(response.content)
+
+        ok_(content['metadata'])
+
+    def test_popcorn_data_exists(self):
+        event = Event.objects.get(title='Test event')
+        event.template.name = 'this is a vid.ly video'
+        event.template.save()
+
+        event.template_environment = {'tag': 'abc123'}
+        event.save()
+
+        edit = PopcornEdit.objects.create(
+            event=event,
+            data={'foo': 'bar'},
+            status=PopcornEdit.STATUS_SUCCESS
+        )
+
+        url = reverse('popcorn:popcorn_data')
+
+        self._login()
+
+        response = self.client.get(url, {'slug': event.slug})
+        eq_(response.status_code, 200)
+
+        content = json.loads(response.content)
+
+        eq_(content['data'], edit.data)
+
+    def test_popcorn_editor(self):
+        event = Event.objects.get(title='Test event')
+        event.template.name = 'this is a vid.ly video'
+        event.template.save()
+
+        event.template_environment = {'tag': 'abc123'}
+        event.save()
+
+        url = reverse('popcorn:render_editor', args=(event.slug,))
+
+        response = self.client.get(url, {'slug': event.slug})
+        eq_(response.status_code, 302)
+
+        event.privacy = Event.PRIVACY_COMPANY
+        event.save()
+
+        response = self.client.get(url, {'slug': event.slug})
+        eq_(response.status_code, 302)
+
+        self._login()
+
+        response = self.client.get(url, {'slug': event.slug})
+        eq_(response.status_code, 200)
--- a/airmozilla/popcorn/views.py
+++ b/airmozilla/popcorn/views.py
@@ -1,5 +1,7 @@
-import requests
+from urlparse import urlparse
+
 from jsonview.decorators import json_view
+import requests
 
 from django import http
 from django.contrib.auth.decorators import login_required
@@ -8,6 +10,7 @@ from django.shortcuts import get_object_or_404, redirect, render
 from airmozilla.main.helpers import thumbnail
 from airmozilla.main.models import Event, VidlySubmission
 from airmozilla.main.views import EventView
+from airmozilla.popcorn.models import PopcornEdit
 
 
 def add_cors_header(value):
@@ -20,18 +23,7 @@ def add_cors_header(value):
     return decorator
 
 
-@add_cors_header('*')
-@json_view
-def event_meta_data(request):
-    slug = request.GET.get('slug')
-    event = get_object_or_404(Event, slug=slug)
-
-    image = event.picture and event.picture.file or event.placeholder_img
-    geometry = '160x90'
-    crop = 'center'
-
-    thumb = thumbnail(image, geometry, crop=crop)
-
+def get_video_url(event):
     if event.template and 'vid.ly' in event.template.name.lower():
         tag = event.template_environment['tag']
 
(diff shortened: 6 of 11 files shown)


airmozilla/popcorn/static/popcorn/js/editor.js
```diff
@@ -1,2 +1,27 @@
...
+
+PopcornEditor.listen(PopcornEditor.events.loaded, function () {
+  $.ajax({
+    method: 'GET',
+    url: '/popcorn/data/',
+    contentType: 'application/json',
+    data: {
+      slug: getSlug()
+    }
+  })
```
peterbe: You can actually just do
```javascript
$.get('/popcorn/data/', {slug: getSlug()})
.done...
```

airmozilla/popcorn/static/popcorn/js/editor.js
```diff
@@ -1,2 +1,27 @@
...
+      re = /\/popcorn\/edit\/(.+)\//,
+      match = pathname.match(re);
+
+  return match[1];
+}
+
+PopcornEditor.listen(PopcornEditor.events.loaded, function () {
+  $.ajax({
+    method: 'GET',
+    url: '/popcorn/data/',
```
peterbe: We need to eradicate all hardcoded URLs from the static files. 
This should only exist has a written string in one single place `popcorn/urls.py`.
The way to "transfer" such a value from django into a script is to use a tag and a `data` attribute.  

Here's one example: https://github.com/mozilla/airmozilla/blob/master/airmozilla/main/templates/main/main_base.html#L169-L175
https://github.com/mozilla/airmozilla/blob/master/airmozilla/main/static/main/js/autocompeter.js#L7-L11

You don't have to use a non-standard tag. Perhaps you can attach...
```
data-editor-url="{{ url('popcorn:data') }}"
```
... on the `body` tag or something.

airmozilla/popcorn/static/popcorn/js/editor.js
```diff
@@ -1,2 +1,27 @@
...
+      slug: getSlug()
+    }
+  })
+  .done(function (response) {
+    if(response.exists) {
+      PopcornEditor.loadInfo(response.data);
+    } else {
+      PopcornEditor.createTemplate(response.data);
+    }
+  });
```
peterbe: This lacks a failure resolver. If you can't think of one at least start with something like this:
```javascript
.fail(function() {
    console.warn("Unable to load popcorn data :(");
    console.error.apply(console, arguments);
})
```
Or, if you don't have time, add an `alert()` because otherwise only people who open the web console will be informed.

airmozilla/popcorn/static/popcorn/js/editor.js
```diff
@@ -1,2 +1,27 @@
...
+    contentType: 'application/json',
+    data: {
+      slug: getSlug()
+    }
+  })
+  .done(function (response) {
+    if(response.exists) {
+      PopcornEditor.loadInfo(response.data);
+    } else {
+      PopcornEditor.createTemplate(response.data);
```
peterbe: This doesn't make sense any more after what we decided.

airmozilla/popcorn/static/popcorn/js/editor.js
```diff
@@ -1,2 +1,27 @@
+function getSlug() {
```
peterbe: This whole file needs to be wrapped in something like `$(function() { ... })` so no javascript magic happens until everything has been loaded. 

And ideally I would use a closure so you don't add `getSlug()` to the `window` global scope. 

You can do this pattern if you like:
```
var PopcornIntegration = (function() {
    function getSlug() {
    }
    return {
        init: function() {
          PopcornEditor.listen(...)
        }
    }
});

$(function() {
    PopcornIntegration.init();
});
```

airmozilla/popcorn/static/popcorn/js/editor.js
```diff
@@ -1,2 +1,27 @@
...
+  var pathname = window.location.pathname,
+      re = /\/popcorn\/edit\/(.+)\//,
+      match = pathname.match(re);
+
+  return match[1];
```
peterbe: This whole function ought to be refactored so it doesn't depend on the URL. 
Where I said to use a `data` attribute to transfer the value of the URL from Django to the javascript via a django template, do the same with the value of the slug.

airmozilla/popcorn/views.py
```diff
@@ -51,11 +42,72 @@ def event_meta_data(request):
...
+    slug = request.GET.get('slug')
+    event = get_object_or_404(Event, slug=slug)
+    pattern = re.compile('(.+)\?')
+
+    base_url = get_video_url(event)
+    matches = pattern.match(base_url)
+    if matches:
+        video_url = matches.group()[:-1]
+    else:
+        video_url = base_url
```
peterbe: Don't use a regex. Use `urlparse.urlparse` and `urlparse.urljoin` to drop the randomish cache-buster thingy.

airmozilla/popcorn/views.py
```diff
@@ -51,11 +42,72 @@ def event_meta_data(request):
...
+        break
+    else:
+        video_url = get_video_url(event)
+        thumb = get_thumbnail(event)
+        data = {
+            "thumbnail": thumb.url,
+            "url": video_url,
+            "title": event.title,
+            "duration": event.duration,
+            "type": "AirMozilla",
```
peterbe: What is this `type`? If it's something that ever reaches users' eyes it should be `Air Mozilla`.

Nolski: This is used within Popcorn Editor and changes how it loads the video data. It refers to the [regex map](https://github.com/Nolski/popcorn-editor/blob/master/PopcornEditor/src/util/mediatypes.js#L9) for loading in different media types

airmozilla/popcorn/views.py
```diff
@@ -51,11 +42,72 @@ def event_meta_data(request):
...
+        }
+
+        return {
+            "exists": False,
+            "data": data,
+        }
+
+    return {
+        "exists": True,
+        "data": data,
```
peterbe: DRY. 

As discussed in IRC, use `response.data` or `response.metadata` as a boolean expression in the ajax callback.

peterbe:
It's great! We just need to fix a couple of nits. In particular to get away from relying on hardcoded URLs and `document.location`

airmozilla/popcorn/static/popcorn/PopcornEditor/src/PopcornEditor.js
```diff
@@ -44,6 +44,85 @@ var PopcornEditor = (function () {
...
+                        "thumbnail": video.thumbnail,
+                        "duration": video.duration
+                    }
+                },
+                "currentTime": 0,
+            }]
+        },
+        "tags": ["popcorn"],
+    }
+    this.loadInfo(data);
```
peterbe: The function implies that it creates stuff but it actually does a load. Perhaps refactor to
`PopcornEditor.loadInfo(PopcornEditor.createTemplate(response.video_metadata))`

airmozilla/popcorn/static/popcorn/PopcornEditor/src/editor/media-gallery-editor.js
```diff
@@ -284,11 +286,11 @@ define( [ "util/lang", "util/uri", "util/xhr", "util/keys", "util/mediatypes", "
     resetInput();
   }
 
-  function onSuccess( data ) {
+  function onSuccess( data, imported ) {
     var el = _GALLERYITEM.cloneNode( true ),
         source = data.source;
 
-    if ( !_media.clipData[ source ] ) {
+    if ( !_media.clipData[ source ]  || imported !== undefined) {
```
peterbe: Why does this matter?
If there's a good reason, leave a comment that explains the necessity to distinguish between imported and non-imported.

airmozilla/popcorn/static/popcorn/js/editor.js
Nolski: Had a chat offline about this. Just a little misunderstanding.

airmozilla/popcorn/static/popcorn/js/editor.js
```diff
@@ -1,2 +1,30 @@
...
+    var config = $('slug');
+    if (config.length) {
+      return config.data('url');
+    } else {
+      return '';
+    }
+  }
+
+  PopcornEditor.listen(PopcornEditor.events.loaded, function () {
+    console.log('slug', getSlug());
```
peterbe: debugging?

airmozilla/popcorn/static/popcorn/js/editor.js
```diff
@@ -1,2 +1,30 @@
-// Initialize the editor with the div id and path to Popcorn Editor.
-PopcornEditor.init('editor', '/static/popcorn/PopcornEditor/editor.html');
+$(document).ready(function () {
+  function getSlug() {
+    var config = $('slug');
+    if (config.length) {
+      return config.data('url');
+    } else {
+      return '';
```
peterbe: Why would this ever happen?

airmozilla/popcorn/templates/popcorn/editor.html
```diff
@@ -23,6 +23,7 @@
 {% endblock %}
 
 {% block full_width_content %}
+<slug data-url="{{ slug }}"></slug>
 <div id="editor"></div>
```
peterbe: You could put that `data-url={}` thing on the `<div id="editor">` element which you already have.

airmozilla/popcorn/tests/test_views.py
```diff
@@ -54,3 +54,42 @@ def mocked_head(url, **options):
...
+        # because we're not logged in
+        eq_(response.status_code, 302)
+
+        self._login()
+
+        response = self.client.get(url, {'slug': event.slug})
+        eq_(response.status_code, 200)
+
+        content = json.loads(response.content)
+        print content
```
peterbe: debugging

airmozilla/popcorn/views.py
```diff
@@ -1,5 +1,6 @@
-import requests
 from jsonview.decorators import json_view
+import requests
+from urlparse import urlparse
```
peterbe: `urlparse` is a standard python module and its import line should go first and separated (by a newline) from the third-party modules.

airmozilla/popcorn/views.py
```diff
@@ -51,11 +42,68 @@ def event_meta_data(request):
 
     assert response.status_code == 302, response.status_code
 
+    purl = urlparse(response.headers['location'])
```
peterbe: Comment missing that explains *why* you're doing this. It's because of the `?t=091u23490-123` thing that vid.ly adds, right? And the reason for doing that is because it prevents popcorn-editor to index the media by its video URL.

airmozilla/popcorn/views.py
```diff
@@ -51,11 +42,68 @@ def event_meta_data(request):
...
+    crop = 'center'
+
+    return thumbnail(image, geometry, crop=crop)
+
+
+@add_cors_header('*')
+@json_view
+def event_meta_data(request):
+    slug = request.GET.get('slug')
+    event = get_object_or_404(Event, slug=slug)
```
peterbe: You're using a `.get()`, which implies that you're OK with that it might become `slug=None`. Then you can't do the next line. The `get_object_or_404`. 

I would change these two lines to:
```python
if not request.GET.get('slug'):
    return http.HttpResponseBadRequest('slug')
event = get_object_or_404(Event, slug=request.GET['slug'])
```

peterbe: One cool feature in tornado (the event driven python web framework) is that you pick things up from the request using `self.get_argument('slug')` and it automatically raises a 400 error for you if that variable is missing or an empty string. And if you don't want that to be required you can do things like: `order_by = self.get_argument('order', 'created')` which sets a default.

airmozilla/popcorn/views.py
```diff
@@ -51,11 +42,68 @@ def event_meta_data(request):
...
+
+    video_url = get_video_url(event)
+
+    thumb = get_thumbnail(event)
+
     return {
         'title': event.title,
         'description': event.short_description or event.description,
```
peterbe: Careful as this might contain HTML tags. There's a helper function somewhere called `unhtml` which you might be interested in.

peterbe: Also, the description can be *very* long. How does Popcorn editor cope with that?

airmozilla/popcorn/views.py
```diff
@@ -51,11 +42,68 @@ def event_meta_data(request):
...
-        'video_url': response.headers['location'],
+        'video_url': video_url,
+    }
+
+
+@json_view
+@login_required
+def popcorn_data(request):
+    slug = request.GET.get('slug')
+    event = get_object_or_404(Event, slug=slug)
```
peterbe: ditto

airmozilla/popcorn/views.py
```diff
@@ -51,11 +42,68 @@ def event_meta_data(request):
...
+            "thumbnail": thumb.url,
+            "url": video_url,
+            "title": event.title,
+            "duration": event.duration,
+            "type": "AirMozilla",
+        }
+
+        return {
+            "metadata": data,
+            "slug": slug,
```
peterbe: Why return what you already clearly knew when you called this view?

airmozilla/popcorn/views.py
```diff
@@ -92,7 +139,8 @@ def get(self, request, slug, form=None, conflict_errors=None):
             return self.cant_edit_event(event, request)
 
         context = {
-            'event': event
+            'event': event,
+            'slug': slug,
```
peterbe: ditto.

airmozilla/popcorn/static/popcorn/PopcornEditor/src/editor/media-gallery-editor.js
Nolski: This prevents you from loading in a video (through the import media dialogue) when it was already loaded in through the API

I'll leave a comment

airmozilla/popcorn/templates/popcorn/editor.html
Nolski: Wouldn't it be a bit more semantic to have the slug be a part of the whole page though? It isn't necessarily specific to the editor element.

airmozilla/popcorn/views.py
Nolski: It will just ellipse after a certain amount of characters. HTML tags should be handled by the js template engine

airmozilla/popcorn/views.py
Nolski: The slug needs to exist within the HTML for the javascript to access which requires it to exist within the context.

airmozilla/popcorn/templates/popcorn/editor.html
peterbe: Yes and no. It's a matter of foresight. Like with a global `styles.css` you *could* add a selector there that is only needed on one page. But there's a big chance you decommission that page in the far future but forget to delete it from `styles.css` and then you have a bloated `styles.css` that future developers get uneasy about cleaning up. That's why I favor using `<style>` tags on individual pages that only need a specific set of rules. (If it gets too big I wrote a thing like `eventedit.css` instead of an inline tag in the HTML)

Then, if I find a need a thing on more than just one page. I can either reuse `eventedit.css` on multiple pages. Or if it gets too many pages, it gets promoted to be included in the global `styles.css`. 

Does that help to explain the issue?

Basically, what if we change how cornpop.js works and nothing no longer needs that `data-slug="..."` in `main_base.html` then `main_base.html` gets scarily bloated.

airmozilla/popcorn/views.py
peterbe: Seems strange but if you say so, let's leave it then. 
"Hey Mike, what's your first name? Mike? Oh cool. Thanks Mike for informing me"  :)

Nolski:
@peterbe once tests pass this should be good to look at again.

airmozilla/popcorn/static/popcorn/PopcornEditor/src/PopcornEditor.js
```diff
@@ -44,6 +44,85 @@ var PopcornEditor = (function () {
...
+                        "thumbnail": video.thumbnail,
+                        "duration": video.duration
+                    }
+                },
+                "currentTime": 0,
+            }]
+        },
+        "tags": ["popcorn"],
+    }
+    return data;
```
peterbe: Yay!

airmozilla/popcorn/static/popcorn/PopcornEditor/src/editor/media-gallery-editor.js
```diff
@@ -700,6 +702,16 @@ define( [ "util/lang", "util/uri", "util/xhr", "util/keys", "util/mediatypes", "
     _searchInput.addEventListener( "keydown", onEnter );
 
     _addBtn.addEventListener( "click", onAddMediaClick );
+
+    Event.listen('load', function (data) {
+      var clipData = data.data.media[0].clipData,
+          clip;
+
+      for (url in clipData) {
```
peterbe: These days you can let yourself use the `clipData.forEach(function(clip) {...` pattern.

airmozilla/popcorn/static/popcorn/js/editor.js
```diff
@@ -1,2 +1,29 @@
-// Initialize the editor with the div id and path to Popcorn Editor.
-PopcornEditor.init('editor', '/static/popcorn/PopcornEditor/editor.html');
+$(document).ready(function () {
+  function getSlug() {
+    var config = $('slug');
+    if (config.length) {
+      return config.data('url');
+    } else {
+      console.error('Slug not provided in DOM. Cannot get slug');
```
peterbe: I would do `throw 'Slug not provided...'`

airmozilla/popcorn/static/popcorn/js/editor.js
```diff
@@ -1,2 +1,29 @@
...
+    }
+  }
+
+  PopcornEditor.listen(PopcornEditor.events.loaded, function () {
+    $.get('/popcorn/data/', {
```
peterbe: This too ought to come from a `data-` tag somewhere. No hardcoded URLs please. URLs should only be "hardcoded" in one place and that's `urls.py`

airmozilla/popcorn/static/popcorn/PopcornEditor/src/editor/media-gallery-editor.js
Nolski: clipData is an `object` not an `Array`

airmozilla/popcorn/static/popcorn/js/editor.js
Nolski: I'm not quite sure how this would be kept in `urls.py` the path to the editor.html file is _strictly_ related to the PopcornEditor.js code. I could hard code it into the template file but that also seems to disconnect it from the actual code that it is semantically linked to.

peterbe: So that URL there relates to [this](https://github.com/Nolski/airmozilla/blob/editor-load-video/airmozilla/popcorn/urls.py#L13-L15) right?
Once you have a URL written down in a `urls.py` you're supposed to refer to it by the name. It's worth it in the long run. 

This line would be replaced with `$('slug').data('data-url')` or something. 
And in the template you'd do something like this:
```html
<slug data-data-url="{{ url('popcorn:data') }}">
```

airmozilla/popcorn/static/popcorn/PopcornEditor/src/editor/media-gallery-editor.js
peterbe: dang it! Good catch. Perhaps put in a comment that explains that for the next person who has the same thought as me.

airmozilla/popcorn/static/popcorn/js/editor.js
```diff
@@ -1,2 +1,29 @@
...
+    .done(function (response) {
+      if(response.data) {
+        PopcornEditor.loadInfo(response.data);
+      } else {
+        PopcornEditor.loadInfo(PopcornEditor.createTemplate(response.metadata));
+      }
+    })
+    .fail(function() {
+      console.warn("Unable to load popcorn data :(");
+      console.error.apply(console, arguments);
```
peterbe: These actually don't help the user that much. Only if they know how to enable the web developer web console will the get a hint that if finished loading but didn't work. 

Couldn't you do something like 
```javascript
.fail(function() {
    ...same console.error lines as before...
    $('#load-error').show();
})
```
And in `editor.html` something like
```html
<div id="load-error" style="display:none">
<p>Someone something something</p>
</div>
```

airmozilla/popcorn/templates/popcorn/editor.html
```diff
@@ -23,6 +23,7 @@
 {% endblock %}
 
 {% block full_width_content %}
+<slug data-url="{{ slug }}"></slug>
```
peterbe: I'm still not convinced we need this extra tag. We should be able to put this into the `<div id="editor">` tag.

peterbe: That means, just being part of the `<div id="editor">` tag it means that those data blocks are related to the editor thing.

airmozilla/popcorn/static/popcorn/js/editor.js
Nolski: Nope that url refers to [this](https://github.com/Nolski/airmozilla/blob/editor-load-video/airmozilla/popcorn/static/popcorn/PopcornEditor/editor.html) which is a static HTML file that belongs to the PopcornEditor

airmozilla/popcorn/static/popcorn/js/editor.js
Nolski: This is on my TODO list but seeing as exporting edits also isn't yet implemented either I kind of put that on the back burner.

airmozilla/popcorn/templates/popcorn/editor.html
Nolski: fair enough

airmozilla/popcorn/static/popcorn/js/editor.js
peterbe: fair enough.

airmozilla/popcorn/static/popcorn/js/editor.js
peterbe: Isn't `/popcorn/data/` a URL in `urls.py` in django?

airmozilla/popcorn/static/popcorn/js/editor.js
```diff
@@ -1,2 +1,29 @@
-// Initialize the editor with the div id and path to Popcorn Editor.
-PopcornEditor.init('editor', '/static/popcorn/PopcornEditor/editor.html');
+$(document).ready(function () {
+  function getSlug() {
+    var config = $('slug');
+    if (config.length) {
+      return config.data('url');
```
peterbe: When you remove `<slug>` and use `<div id="editor">` instead, change the name from `url` to `slug`.

peterbe:
Ping me in IRC so we can figure out why your schemamigrations stuff doesn't want to generate a new migration file.

Nolski:
@peterbe Looks like I had to drop the test table to get tests to pass (also needed to add the database migration)

airmozilla/popcorn/static/popcorn/PopcornEditor/src/editor/media-gallery-editor.js
peterbe: Don't forget that comment. I think it's an important underlying design thing to understand.

airmozilla/popcorn/static/popcorn/js/editor.js
```diff
@@ -1,2 +1,20 @@
-// Initialize the editor with the div id and path to Popcorn Editor.
-PopcornEditor.init('editor', '/static/popcorn/PopcornEditor/editor.html');
+$(document).ready(function () {
+  PopcornEditor.listen(PopcornEditor.events.loaded, function () {
+    $.get($('#editor').data('url'), {
+        slug: $('#editor').data('slug')
+    })
```
peterbe: Awesome! Keeps it simple.

airmozilla/popcorn/templates/popcorn/editor.html
```diff
@@ -23,7 +23,7 @@
 {% endblock %}
 
 {% block full_width_content %}
-<div id="editor"></div>
+<div id="editor" data-slug="{{ slug }}" data-url="{{ url(popcorn:data) }}"></div>
```
peterbe: How could this have worked in testing? Perhaps we forgot to add a test that executes this view. 
It's supposed to be `{{ url('popcorn:data') }}`. 

Check your coverage first and then, write a test that executes this template (which should fail because the expression `{{ url(popcorn:data) }}` is most likely not valid in Jinja.

airmozilla/popcorn/tests/test_views.py
```diff
@@ -54,3 +54,41 @@ def mocked_head(url, **options):
...
+        event = Event.objects.get(title='Test event')
+        event.template.name = 'this is a vid.ly video'
+        event.template.save()
+
+        event.template_environment = {'tag': 'abc123'}
+        event.save()
+
+        url = reverse('popcorn:popcorn_data')
+
+        response = self.client.get(url, {'slug': event.slug})
```
peterbe: Before this do...
```python
response = self.client.get(url)
# bad request because it's missing the slug
eq_(response.status_code, 400)

response = self.client.get(url, {'slug': ''})
# still not a valid slug value
eq_(response.status_code, 400)
```

peterbe:
Seems to lack test coverage in a couple of places. 
The rendering of the page that has `<div id="editor" data-slug="{{ slug }}">` might deserve its own test. 

Also you have no test coverage of when there is 1 or more PopcornEdits. E.g. that for-loop with the break in it. 

If you can't remember how the coverage was done, see [the readme](https://github.com/mozilla/airmozilla#tests-and-test-coverage).

I'm really eager to see this land. We are so close! I'll help by building a little admin thingy so you can get insights into known edits.

Nolski:
@peterbe squashed and rebased.

PR #415, opened 2015-08-02
Popcoder integration

Nolski:
This includes:
- Code for running a cronjob which will render the first of all video submissions
- Webhook for updating our database when Vidly finishes rendering our new video
- Tests for the webhook.

### DIFF
--- a/airmozilla/manage/tests/views/test_vidlymedia.py
+++ b/airmozilla/manage/tests/views/test_vidlymedia.py
@@ -26,8 +26,7 @@ from airmozilla.manage.tests.test_vidly import (
 )
 from .base import ManageTestCase
 
-SAMPLE_MEDIA_SUBMITTED_XML = """
-<?xml version="1.0"?>
+SAMPLE_MEDIA_SUBMITTED_XML = """<?xml version="1.0"?>
 <Response>
 <Message>All medias have been added.</Message>
 <MessageCode>2.1</MessageCode>
@@ -81,7 +80,7 @@ SAMPLE_MEDIA_RESULT_FAILED = """
     </Task>
   </Result>
 </Response>
-"""
+""".strip()
 
 SAMPLE_MEDIA_RESULT_SUCCESS = """
 <?xml version="1.0"?>
@@ -125,7 +124,7 @@ SAMPLE_MEDIA_RESULT_SUCCESS = """
         </Task>
     </Result>
 </Response>
-"""
+""".strip()
 
 
 class TestVidlyMedia(ManageTestCase):
--- a/airmozilla/popcorn/renderer.py
+++ b/airmozilla/popcorn/renderer.py
@@ -1,0 +1,107 @@
+import os
+import tempfile
+import time
+from uuid import uuid4
+
+import boto
+from boto.s3.key import Key
+from funfactory.urlresolvers import reverse
+from popcoder.popcoder import process_json
+
+from django.db import transaction
+from django.conf import settings
+from django.utils import timezone
+
+from airmozilla.base.utils import build_absolute_url, prepare_vidly_video_url
+from airmozilla.popcorn.models import PopcornEdit
+from airmozilla.main.models import Event, VidlySubmission
+from airmozilla.manage import vidly
+from airmozilla.uploads.models import Upload
+
+
+@transaction.atomic
+def render_edit(edit_id):
+    edit = PopcornEdit.objects.get(id=edit_id)
+    event = edit.event
+
+    filename = '%s.webm' % edit.id
+    filepath = os.path.join(tempfile.gettempdir(), filename)
+    if not (os.path.isfile(filepath) and
+            os.stat(filepath)[os.stat.st_size] > 0):
+        edit.status = PopcornEdit.STATUS_PROCESSING
+        edit.save()
+        process_json(edit.data, out=filepath)
+
+    upload_file_name = '%s.webm' % uuid4().hex[:13]
+    # Uploads file to s3
+    connection = boto.connect_s3(
+        settings.AWS_ACCESS_KEY_ID,
+        settings.AWS_SECRET_ACCESS_KEY,
+    )
+
+    nonexistent = connection.lookup(settings.S3_UPLOAD_BUCKET)
+    if nonexistent:
+        bucket = connection.create_bucket(settings.S3_UPLOAD_BUCKET)
+    else:
+        bucket = connection.get_bucket(settings.S3_UPLOAD_BUCKET)
+
+    video_key = Key(bucket)
+
+    video_key.key = upload_file_name
+    start = time.time()
+    video_key.set_contents_from_filename(filepath)
+    end = time.time()
+
+    video_url = video_key.generate_url(expires_in=0, query_auth=False)
+    video_url = prepare_vidly_video_url(video_url)
+
+    filesize = os.stat(filepath).st_size
+
+    Upload.objects.create(
+        user=edit.user,
+        url=video_url,
+        file_name=upload_file_name,
+        mime_type='video/webm',
+        size=filesize,
+        upload_time=int(end - start)
+    )
+
+    webhook_url = build_absolute_url(reverse('popcorn:vidly_webhook'))
+    token_protection = event.privacy != Event.PRIVACY_PUBLIC
+
+    tag, error = vidly.add_media(
+        url=video_url,
+        token_protection=token_protection,
+        hd=True,
+        notify_url=webhook_url,
+    )
+
+    VidlySubmission.objects.create(
+        event=event,
+        url=video_url,
+        tag=tag,
+        hd=True,
+        submission_error=error
+    )
+
+    # raise exception if error
+    if error:
+        raise Exception(error)
+
+    edit.status = PopcornEdit.STATUS_SUCCESS
+    edit.finished = timezone.now()
+    edit.save()
+
+    os.remove(filepath)
+
+
+def render_all_videos(verbose=False):
+    # Re process if start time is more than 2 hours
+    pending_edits = PopcornEdit.objects.filter(PopcornEdit.STATUS_PENDING)
+    if pending_edits.count() > 1:
+        if verbose:
+            print "Currently more than one edit. Skipping for now"
+            print pending_edits.count()
+    pending_edits = pending_edits[:1]
+    for pending_edit in pending_edits:
+        render_edit(pending_edit.id)
--- a/airmozilla/popcorn/tests/test_renderer.py
+++ b/airmozilla/popcorn/tests/test_renderer.py
@@ -1,0 +1,88 @@
+from cStringIO import StringIO
+import mock
+from mock import Mock
+
+from nose.tools import eq_, ok_
+
+from airmozilla.base.tests.testbase import DjangoTestCase
+from airmozilla.main.models import Event, VidlySubmission
+from airmozilla.manage.tests.views.test_vidlymedia import (
+    SAMPLE_MEDIA_SUBMITTED_XML,
+)
+from airmozilla.popcorn.renderer import render_edit
+from airmozilla.popcorn.models import PopcornEdit
+
+
+class TestPopcornRender(DjangoTestCase):
+    fixtures = ['airmozilla/manage/tests/main_testdata.json']
+
+    @mock.patch('airmozilla.manage.vidly.urllib2')
+    @mock.patch('airmozilla.popcorn.renderer.Key')
+    @mock.patch('boto.connect_s3')
+    @mock.patch('airmozilla.popcorn.renderer.process_json')
+    def test_render(self, p_process_json, p_connect_s3, p_key, p_urllib2):
+        def make_mock_request(url, querystring):
+            return mock.MagicMock()
+
+        def mocked_urlopen(request):
+            xml_string = SAMPLE_MEDIA_SUBMITTED_XML
+            return StringIO(xml_string)
+
+        def process_json_mock(data, out, background_color='#000000'):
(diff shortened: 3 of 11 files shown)


airmozilla/manage/tests/views/test_vidlymedia.py
```diff
@@ -83,8 +81,7 @@
 </Response>
 """
 
-SAMPLE_MEDIA_RESULT_SUCCESS = """
-<?xml version="1.0"?>
+SAMPLE_MEDIA_RESULT_SUCCESS = """<?xml version="1.0"?>
```
peterbe: All of these changes were unnecessary. Note how there's a `""".strip()` at the end. This way the first line can be after a nice little newline.

airmozilla/popcorn/renderer.py
```diff
@@ -0,0 +1,92 @@
...
+    tag, error = vidly.add_media(
+        prepare_vidly_video_url(video_url),
+        token_protection=token_protection,
+        hd=True,
+        notify_url=webhook_url,
+    )
+
+    # raise exception if error
+    if error:
+        raise "Vidly upload failed"
```
peterbe: Invalid Python. You can just do...:

```python
if error:
    raise Exception(error)
```

airmozilla/popcorn/renderer.py
```diff
@@ -0,0 +1,92 @@
...
+
+    VidlySubmission.objects.create(
+        event=event,
+        url=video_url,
+        tag=tag,
+        hd=True,
+        submission_error=error or None
```
peterbe: Can never be anything other than None

airmozilla/popcorn/renderer.py
```diff
@@ -0,0 +1,92 @@
...
+        mime_type='video/webm',
+        size=filesize,
+        upload_time=timezone.now()
+    )
+
+    webhook_url = build_absolute_url(reverse('popcorn:vidly_webhook'))
+    token_protection = event.privacy != Event.PRIVACY_PUBLIC
+
+    tag, error = vidly.add_media(
+        prepare_vidly_video_url(video_url),
```
peterbe: The URL we pass to `vidly.add_media` need to match what we pass to `VidlySubmission.objects.create(url=video_url, ...`
So, change it to this:
```python
video_url = prepare_vidly_video_url(video_url)
tag, error = vidly.add_media(
    video_url,
    ...
)
VidlySubmission.objects.create(
    event=event,
    url=video_url,
    ...
)
```

airmozilla/popcorn/renderer.py
```diff
@@ -0,0 +1,92 @@
...
+        token_protection=token_protection,
+        hd=True,
+        notify_url=webhook_url,
+    )
+
+    # raise exception if error
+    if error:
+        raise "Vidly upload failed"
+
+    VidlySubmission.objects.create(
```
peterbe: Actually, let's move this to the `)` right after the `vidly.add_media(...)`. 
That way, you can keep the `submission_error=error`

airmozilla/popcorn/tests/test_views.py
```diff
@@ -147,3 +157,140 @@ def test_popcorn_editor(self):
...
+        VidlySubmission.objects.create(
+            event=event,
+            url=file_url,
+            tag=tag,
+            token_protection=False
+        )
+
+        response = self.client.post(url, {'xml': xml_string})
+        eq_(response.status_code, 200)
+        eq_('OK\n', response.content)
```
peterbe: You might want to check that the `VidlySubmission` that was created on line 192 changed. 
E.g. First change line 192 to `vidly_submission = VidlySubmission.objects.create(`
then do something like this:
```python
vidly_submission = VidlySubmission.objects.get(id=vidly_submission.id)
ok_(vidly_submission.finished)
```

airmozilla/popcorn/tests/test_views.py
```diff
@@ -147,3 +157,140 @@ def test_popcorn_editor(self):
...
+
+    @mock.patch('airmozilla.manage.vidly.urllib2')
+    def test_vidly_webhook_input_error(self, p_urllib2):
+        def make_mock_request(url, querystring):
+            return mock.MagicMock()
+
+        def mocked_urlopen(request):
```
peterbe: Does this ever get called?

airmozilla/popcorn/tests/test_views.py
```diff
@@ -147,3 +157,140 @@ def test_popcorn_editor(self):
...
+            token_protection=False
+        )
+
+        response = self.client.post(url, {'xml': xml_string})
+        eq_(response.status_code, 200)
+        eq_('OK\n', response.content)
+
+    @mock.patch('airmozilla.manage.vidly.urllib2')
+    def test_vidly_webhook_input_error(self, p_urllib2):
+        def make_mock_request(url, querystring):
```
peterbe: here too. I'm not sure this ever gets used.

airmozilla/popcorn/renderer.py
```diff
@@ -0,0 +1,92 @@
...
+
+    # Uploads file to s3
+    connection = boto.connect_s3(
+        settings.AWS_ACCESS_KEY_ID,
+        settings.AWES_SECRET_ACCESS_KEY,
+    )
+    bucket = connection.get_bucket('vidly_uploads')
+    video_key = Key(bucket)
+
+    video_key.key = filename
```
peterbe: This needs to change. Either here or on line 25. 
We need something random. 
Otherwise you'd be able to guess that the URL is `https://s3.amazon.com/airmo/1.webm` or `https://s3.amazon.com/airmo/<SOME INTEGER>.webm` and you'd be able to just `wget` integers until you luckily manage to download a private video that you shouldn't be able to `wget`.

peterbe: Now I worry that I have other code that makes the same mistake.

airmozilla/popcorn/renderer.py
```diff
@@ -0,0 +1,92 @@
...
+
+@transaction.atomic
+def render_edit(id):
+    edit = PopcornEdit.objects.get(id=id)
+    event = edit.event
+
+    filename = '%s.webm' % edit.id
+    filepath = os.path.join(tempfile.gettempdir(), filename)
+    filesize = os.stat(filepath)[os.stat.st_size]
+    if not (os.path.isfile(filepath) and filesize > 0):
```
peterbe: These two lines don't make sense. 
It's like saying:
```
What is your height?
If you exist and your height is more than zero: ...
```

peterbe: In other words, suppose the file does NOT exist, then `os.stat(filepath)[os.stat.st_size]` should fail.

airmozilla/popcorn/tests/test_views.py
```diff
@@ -1,11 +1,21 @@
+from cStringIO import StringIO
+import mock
+import json
```
peterbe: FYI, the correct pattern is:
```
pythonstandardlib
pythonstandardlib
pythonstandardlib

thirdpartypythonlibs
thirdpartypythonlibs
thirdpartypythonlibs

djangoimports
djangoimports
djangoimports

thirdpartydjangospecificimports
thirdpartydjangospecificimports
thirdpartydjangospecificimports

localimports
localimports
localimports
```

airmozilla/popcorn/urls.py
```diff
@@ -13,4 +13,7 @@
     url(r'^data/$',
         views.popcorn_data,
         name='popcorn_data'),
+    url(r'^vidly_webhook$',
```
peterbe: Try to avoid `_` underscore characters in the URL part. It's OK, and encouraged, in the function names.

airmozilla/popcorn/views.py
```diff
@@ -106,6 +113,67 @@ def popcorn_data(request):
...
+        task = struct['Response']['Result']['Task']
+
+        try:
+            migrate_submission(task)
+        except VidlySubmission.DoesNotExist:
+            # remember, we can't trust the XML since it's publicly
+            # available and exposed as a webhook
+            response = http.HttpResponse('Vidly Submission does not exist')
+            response.status_code = 404
+            return response
```
peterbe: These three lines can be replaced with 
```
raise http.Http404('Vidly Submission does not exist')
```

peterbe: scratch this. See note above about using `get_object_or_404`

airmozilla/popcorn/tests/test_views.py
Nolski: yeah, `vidly.query` calls this

airmozilla/popcorn/tests/test_views.py
Nolski: see above

Nolski:
@peterbe if the builds pass I think that I've addressed all of your comments.

airmozilla/popcorn/renderer.py
```diff
@@ -0,0 +1,101 @@
...
+
+    filename = '%s.webm' % edit.id
+    filepath = os.path.join(tempfile.gettempdir(), filename)
+    if not (os.path.isfile(filepath) and
+            os.stat(filepath)[os.stat.st_size] > 0):
+        edit.status = PopcornEdit.STATUS_PROCESSING
+        edit.save()
+        process_json(edit.data, out=filepath)
+
+    upload_file = uuid4().hex[:13]
```
peterbe: wanna call that `'%s.webm' % uuid4().hex[:13]` instead so the filename makes sense when it's on S3?

airmozilla/popcorn/renderer.py
```diff
@@ -0,0 +1,101 @@
...
+    # Uploads file to s3
+    connection = boto.connect_s3(
+        settings.AWS_ACCESS_KEY_ID,
+        settings.AWS_SECRET_ACCESS_KEY,
+    )
+    bucket = connection.get_bucket('vidly_uploads')
```
peterbe: Instead of a hardcoded string this needs to be `settings.S3_UPLOAD_BUCKET`

peterbe: Also, can you change this to some sort of `get_or_create_bucket()` code? It might not exist on first run.

airmozilla/popcorn/renderer.py
peterbe: The variable `upload_file` isn't great. 
Is it a name or a path? Or is it an instance of a Upload class? See what I mean?

airmozilla/popcorn/renderer.py
```diff
@@ -0,0 +1,101 @@
...
+        event=event,
+        url=video_url,
+        tag=tag,
+        hd=True,
+        submission_error=error
+    )
+
+    # raise exception if error
+    if error:
+        raise Exception("Vidly upload failed")
```
peterbe: Perhaps instead do `raise Exception(error)` so that the error message contains information about what went wrong with vidly. Alternatively you'd have to look up the `VidlySubmission` object and read its `submission_error` thing.

airmozilla/popcorn/renderer.py
Nolski: Shouldn't it exist by default? [this code](https://github.com/mozilla/airmozilla/blob/master/airmozilla/manage/views/events.py#L671) seems to just use it without creating a bucket

airmozilla/popcorn/tests/test_renderer.py
```diff
@@ -0,0 +1,86 @@
...
+        event = Event.objects.get(title='Test event')
+
+        edit = PopcornEdit.objects.create(
+            event=event,
+            status=PopcornEdit.STATUS_PENDING,
+            data={},
+            user=event.creator,
+        )
+
+        render_edit(edit.id)
```
peterbe: No checking afterwards? Like checking that there is now a vidly submission etc?

airmozilla/popcorn/views.py
```diff
@@ -106,6 +113,60 @@ def popcorn_data(request):
...
+        struct = xmltodict.parse(xml_string)
+    except ExpatError:
+        return http.HttpResponseBadRequest("Bad 'xml'")
+
+    try:
+        task = struct['Response']['Result']['Task']
+        migrate_submission(task)
+    except KeyError:
+        # If it doesn't have a "Result" or "Task", it was just a notification
+        # that the media was added.
```
peterbe: This is misleading. A `KeyError` could happen on line 131 and/or line 132. You just don't know. It's bad to put too broad `try:except` blocks on things. What if you have a bug in `migrate_submission()` that raises a `KeyError` sometimes then we'd want that to crash and burn so we can get alerted and fix it.

airmozilla/popcorn/views.py
```diff
@@ -106,6 +113,60 @@ def popcorn_data(request):
...
+
+    xml_string = request.POST['xml'].strip()
+    try:
+        struct = xmltodict.parse(xml_string)
+    except ExpatError:
+        return http.HttpResponseBadRequest("Bad 'xml'")
+
+    try:
+        task = struct['Response']['Result']['Task']
+        migrate_submission(task)
```
peterbe: I still don't like the naming here. If it was called `migrate_user` you'd naturally thing the first argument into that function is a `user` thing. Not a `task`. I recommend you make the `migrate_submission` function oblivious of tasks. E.g. 
```python
migrate_submission(get_object_or_404(
    VidlySubmission,
    url=task['SourceFile'],
    tag=task['MediaShortLink']
))
```
And then change `migrate_submission` to something like this:
```python
def migrate_submission(submission):
    results = vidly.query(submission.tag)
    ...
```

peterbe:
r+

Nolski:
rebased and squashed @peterbe

PR #419, opened 2015-08-05
Hooked up save button for creating PopcornEdit

Nolski:
Includes tests!

### DIFF
--- a/airmozilla/popcorn/static/popcorn/js/editor.js
+++ b/airmozilla/popcorn/static/popcorn/js/editor.js
@@ -17,4 +17,21 @@ $(document).ready(function () {
   })
   // Initialize the editor with the div id and path to Popcorn Editor.
   PopcornEditor.init('editor', '/static/popcorn/PopcornEditor/editor.html');
+  PopcornEditor.listen('save', function (message) {
+    $.post($('#editor').data('save'), {
+      data: JSON.stringify(message.data),
+      slug: $('#editor').data('slug'),
+      csrfmiddlewaretoken: $('input[name="csrfmiddlewaretoken"]').val()
+    })
+    .done(function (response) {
+      var message = "Your edit has been saved and sent it for transcoding. "
+      + "Once it completes, which can take several minutes, we'll automatically"
+      + " update the event to use your latest video edits.";
+      alert(message);
+    })
+    .fail(function () {
+      console.warn('Unable to save edit :(');
+      console.error.apply(console, arguments);
+    });
+  });
 });
--- a/airmozilla/popcorn/tests/test_views.py
+++ b/airmozilla/popcorn/tests/test_views.py
@@ -157,6 +157,49 @@ class TestPopcornEvent(DjangoTestCase):
         response = self.client.get(url, {'slug': event.slug})
         eq_(response.status_code, 200)
 
+    def test_save_edit(self):
+        event = Event.objects.get(title='Test event')
+        event.template.name = 'this is a vid.ly video'
+        event.template.save()
+
+        event.template_environment = {'tag': 'abc123'}
+        event.save()
+
+        url = reverse('popcorn:save_edit')
+
+        self._login()
+
+        response = self.client.post(url, {
+            'slug': event.slug,
+        })
+        # Should error due to missing data field
+        eq_(response.status_code, 400)
+
+        response = self.client.post(url, {
+            'data': '{}',
+        })
+        # Should error due to missing slug field
+        eq_(response.status_code, 400)
+
+        response = self.client.post(url, {
+            'slug': 'does_not_exist',
+            'data': '{}',
+        })
+        eq_(response.status_code, 404)
+
+        response = self.client.post(url, {
+            'slug': event.slug,
+            'data': '{}',
+        })
+
+        edit = PopcornEdit.objects.get(
+            event=event,
+            status=PopcornEdit.STATUS_PENDING,
+        )
+
+        eq_(response.status_code, 200)
+        eq_(edit.id, json.loads(response.content)['id'])
+
     @mock.patch('airmozilla.manage.vidly.urllib2')
     def test_vidly_webhook(self, p_urllib2):
         xml_string = SAMPLE_MEDIA_RESULT_SUCCESS
--- a/airmozilla/popcorn/views.py
+++ b/airmozilla/popcorn/views.py
@@ -1,3 +1,4 @@
+import json
 from urlparse import urlparse
 from xml.parsers.expat import ExpatError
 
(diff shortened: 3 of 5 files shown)


airmozilla/popcorn/static/popcorn/js/editor.js
```diff
@@ -17,4 +17,20 @@ $(document).ready(function () {
   })
   // Initialize the editor with the div id and path to Popcorn Editor.
   PopcornEditor.init('editor', '/static/popcorn/PopcornEditor/editor.html');
+  PopcornEditor.listen('save', function (message) {
+    console.log(message);
```
peterbe: debug?

airmozilla/popcorn/static/popcorn/js/editor.js
```diff
@@ -17,4 +17,20 @@ $(document).ready(function () {
...
+      data: JSON.stringify(message.data),
+      slug: $('#editor').data('slug'),
+      csrfmiddlewaretoken: $('input[name="csrfmiddlewaretoken"]').val()
+    })
+    .done(function (response) {
+      alert('Event is saved and processing with PopcornEdit.id = '
+           + response.id);
+    })
+    .fail(function (response) {
+      console.warn('Unable to save edit :(')
```
peterbe: semi colon

airmozilla/popcorn/static/popcorn/js/editor.js
```diff
@@ -17,4 +17,20 @@ $(document).ready(function () {
...
+    $.post($('#editor').data('save'), {
+      data: JSON.stringify(message.data),
+      slug: $('#editor').data('slug'),
+      csrfmiddlewaretoken: $('input[name="csrfmiddlewaretoken"]').val()
+    })
+    .done(function (response) {
+      alert('Event is saved and processing with PopcornEdit.id = '
+           + response.id);
+    })
+    .fail(function (response) {
```
peterbe: we don't use the `response`

airmozilla/popcorn/tests/test_views.py
```diff
@@ -157,6 +157,47 @@ def test_popcorn_editor(self):
...
+        event.save()
+
+        url = reverse('popcorn:save_edit')
+
+        self._login()
+
+        response = self.client.post(url, {
+            'slug': 'does_not_exist',
+        })
+        eq_(response.status_code, 400)
```
peterbe: Shouldn't that be 404?

peterbe: oh. I see. It's because `data` is missing. Please add a comment that this 400 is to be expected because there is no `data`.

Nolski: It checks for the field in the request before it get's the event by the slug. I'll change to event.slug to be more clear

airmozilla/popcorn/static/popcorn/js/editor.js
```diff
@@ -17,4 +17,20 @@ $(document).ready(function () {
...
+  PopcornEditor.listen('save', function (message) {
+    console.log(message);
+    $.post($('#editor').data('save'), {
+      data: JSON.stringify(message.data),
+      slug: $('#editor').data('slug'),
+      csrfmiddlewaretoken: $('input[name="csrfmiddlewaretoken"]').val()
+    })
+    .done(function (response) {
+      alert('Event is saved and processing with PopcornEdit.id = '
+           + response.id);
```
peterbe: This message isn't very user-friendly. Perhaps make it slightly more human before we forget to get around to fixing it with a more fancy modal or something.

airmozilla/popcorn/tests/test_views.py
```diff
@@ -157,6 +157,47 @@ def test_popcorn_editor(self):
...
+
+        response = self.client.post(url, {
+            'slug': 'does_not_exist',
+            'data': '{}',
+        })
+        eq_(response.status_code, 404)
+
+        response = self.client.post(url, {
+            'slug': event.slug,
+            'data': '{}',
```
peterbe: Hmm... If you convert that from a string to a python dict, you should get trapped in `if not data: return BadRequest...`

airmozilla/popcorn/views.py
```diff
@@ -113,6 +113,31 @@ def popcorn_data(request):
...
+@login_required
+@require_POST
+def save_edit(request):
+    slug = request.POST.get('slug')
+    if not slug:
+        return http.HttpResponseBadRequest('slug')
+    data = request.POST.get('data')
+    if not data:
+        return http.HttpResponseBadRequest('data')
+
```
peterbe: I think we should check here that this string can be converted from a JSON string to a python dict object. 
And if it's empty (e.g. `{}`) we should not accept it.

airmozilla/popcorn/tests/test_views.py
Nolski: Why is that? We don't validate the popcorn json when we create a PopcornEdit. I suppose we could do that but it would take quite a lot of work to do it beforehand rather during the edit.

airmozilla/popcorn/static/popcorn/js/editor.js
Nolski: Any ideas? I think having an alert saying your data has been saved is less descriptive and really this will need a fancy modal for it to really be usable. I wanted to keep it out of this PR just so we could troubleshoot the whole process as soon as possible.

airmozilla/popcorn/tests/test_views.py
peterbe: We should check that the string can be converted to a python dict. 
I think the `PopcornEdit.data` field is a JSONField which should hopefully barf but the error might then become an Internal Server Error rather than a 400 Bad Request.

peterbe: We don't want to validate its **content**. Just that it's valid JSON.

airmozilla/popcorn/views.py
Nolski: That's one step of validation but `{}` is no more valid than `{'foo': 'bar'}` so it seems kind of pointless to me.

airmozilla/popcorn/tests/test_views.py
Nolski: I see. I'll update that then.

airmozilla/popcorn/static/popcorn/js/editor.js
peterbe: Oh, I was thinking we'll stick with an `alert` but a mortal don't understand what `PopcordEdit.id = 123` means. 
How about
`alert("Your edit has been saved and sent it for transcoding. Once it completes, which can take several minutes, we'll automatically update the event to use your latest video edits.")`

Nolski: Sure, I'm fine with updating it to that but do we plan on any mortals even being able to see this before we implement the modal?

airmozilla/popcorn/tests/test_views.py
peterbe: Someone might post in `{` which is not valid.

peterbe:
r+ 
Well done!

Nolski:
@peterbe looks like the tests pass this time. Should be good to merge.

PR #422, opened 2015-08-06
Bug-1186669 Re-enable selenium tests

lcamacho:
@peterbe please review


peterbe:
Same old error :(

lcamacho:
Weird things of selenium, on local if executed only selenium tests is ok but if executed the complete battery of tests then fails, which is what occurs on travis. The fail is that on first case calendar.js detect 'America/Los_Angeles' tz but latter case detect the right timezone which is UTC, no idea of how this can happen.

lcamacho:
@peterbe please trigger a new build

peterbe:
Re-opening does that.

lcamacho:
weird fails, not related to calendar, keep appearing on travis

lcamacho:
green light

peterbe:
What a trip! I'm snowed under at the moment on a different project and can't spend super much attention on this but I'm glad it's getting greener.

PR #437, opened 2015-08-11
Fixes Bug 1191415 - Refactor editor.js to use 4 spaces for indentation

anu7495:


peterbe:
@Nolski looks good?

Nolski:
Yeah, I was working with another on a similar PR but this one seems to do the job just fine. :shipit:

PR #487, opened 2015-09-13
Add a 1.25X playback rate option

ehsan:
YouTube supports this playback rate, and it seems like it works
much better than 1.5X when listening to someone speak, in that at
1.5X, what they are saying starts to get hard to understand in a
lot of cases.


peterbe:
Is there any reason to keep the 2x rate? If 1.5 is even too much, 2x must be nuts.

ehsan:
> Is there any reason to keep the 2x rate? If 1.5 is even too much, 2x must be nuts.

Agreed, see #488.

PR #549, opened 2015-10-22
Fix bug 1216825 - Added share links

lcamacho:


lcamacho:
@peterbe r?

airmozilla/main/templates/main/event.html
```diff
@@ -334,6 +334,22 @@ <h1 class="entry-title">
         </div>
       {% endif %}
 
+      <div class="tab">
```
peterbe: This whole thing need to be wrapped in a `{% if event.is_public() %}` block.

peterbe:
It looks a bit weird in a new tab since it's meant to be a pop-up
screenshot 2015-10-22 12 02 46

Perhaps we should consider this: http://jsfiddle.net/stichoza/eyxtj/

peterbe:
Actually, the comment above is applicable to all three options.

lcamacho:
@peterbe this new version should work

airmozilla/main/templates/main/event.html
```diff
@@ -334,6 +335,24 @@ <h1 class="entry-title">
         </div>
       {% endif %}
 
+      {% if event.is_public() %}
+        <div class="tab">
+          <input type="radio" id="tab-share" name="tab-group-1">
+          <label for="tab-share">Share</label>
+          <div class="content">
+            <div>
+              <a class="share" href="https://www.facebook.com/sharer/sharer.php?u={{ make_absolute(url('main:event', event.slug)) }}" target="_blank">Share on Facebook</a>
```
peterbe: The URL should be `urlencode`'ed too. Most characters in it are safe but it's best practice to URL encode them too.

airmozilla/main/templates/main/event.html
```diff
@@ -334,6 +335,24 @@ <h1 class="entry-title">
         </div>
       {% endif %}
 
+      {% if event.is_public() %}
+        <div class="tab">
+          <input type="radio" id="tab-share" name="tab-group-1">
+          <label for="tab-share">Share</label>
+          <div class="content">
+            <div>
```
peterbe: Why use individual `<div>` tags and not `<p>`? The paragraph tag feels more semantic. Is there a particular reason why `<div>` is better?

lcamacho: ``<p>`` left too much space between links nothing else

airmozilla/main/templates/main/event.html
lcamacho: From here https://docs.djangoproject.com/en/1.8/ref/templates/builtins/#url any special character should be encoded

airmozilla/main/templates/main/event.html
lcamacho: I mean this
captura de pantalla 2015-10-22 a las 16 34 35

airmozilla/main/templates/main/event.html
peterbe: We're not using that built-in and the `url()` function will return something like '/foo' which is fine. The the `make_absolute_url('/foo')` will return `http://localhost:8000/foo` which is also fine but that means that the final HTML becomes `href="https://www.facebook.com/sharer/sharer.php?u=http://localhost:8000/foo"` when instead it should be `href="https://www.facebook.com/sharer/sharer.php?u=http%3A%2F%2Flocalhost%3A8000%2Ffoo"`

airmozilla/main/templates/main/event.html
peterbe: that's probably better. The day we have a 4th or 5th or Nth link, we'll have to consider changing to...
```
<p>
    <a ...>...</a><br>
    <a ...>...</a><br>
    <a ...>...</a><br>

    <a ...>...</a><br>
    <a ...>...</a><br>
    <a ...>...</a><br>
</p>
```

lcamacho:
@peterbe r?

peterbe:
THANK YOU AGAIN!

PR #579, opened 2016-01-25
fixes bug 1241529 - Drop peep and use pip 8

peterbe:


bin/update/update.py
```diff
@@ -39,8 +39,9 @@ def update_code(ctx, tag):
         )
         execfile(activate_env, dict(__file__=activate_env))
 
-        ctx.local('%s/bin/pip install bin/peep-2.4.1.tar.gz' % venv_path)
-        ctx.local('%s/bin/peep install -r requirements.txt' % venv_path)
+        # this makes sure we have pip 8.x at least
+        ctx.local('%s/bin/pip install -U pip' % venv_path)
```
erikrose: Here's where you'd use pipstrap, if you like.

peterbe: Done. Almost. https://github.com/mozilla/airmozilla/pull/584

PR #616, opened 2016-02-10
Changes to fix the Bug 1201143

64lines:
This change is for consolidating the STOPWORDS list according to the 1201143 bug tracker.

### DIFF
--- a/airmozilla/base/utils.py
+++ b/airmozilla/base/utils.py
@@ -23,6 +23,20 @@ from django.contrib.staticfiles.storage import staticfiles_storage
 from airmozilla.base.akamai_token_v2 import AkamaiToken
 
 
+STOPWORDS = (
+    "a able about across after all almost also am among an and "
+    "any are as at be because been but by can cannot could dear "
+    "did do does either else ever every for from get got had has "
+    "have he her hers him his how however i if in into is it its "
+    "just least let like likely may me might most must my "
+    "neither no nor not of off often on only or other our own "
+    "rather said say says she should since so some than that the "
+    "their them then there these they this tis to too twas us "
+    "wants was we were what when where which while who whom why "
+    "will with would yet you your".split()
+)
+
+
 def roughly(number, variance_percentage=20):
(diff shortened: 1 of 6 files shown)


peterbe:
Hi @64lines! Welcome to the community. 

As you can see, you have some pyflakes errors. Do you see that?

One way to prevent this from happening is to first run `pyflakes` and `pep8` on the files before you commit. One way to do that is to use `flake8`. (`pip install flake8`). 
You can set up a git pre-commit hook like this:
1) pip install flake8
2) Download https://gist.github.com/peterbe/0fab292e657d003533b9
3) Save it as `.git/hooks/pre-commit` in the repo root directory

64lines:
Hi @peterbe, 

Thank you for the advice! I'm running the command right now and fixing the possible problems.

peterbe:
If you run `flake8` on that `search/utils.py` page it'll tell you what's wrong.

peterbe:
We can ignore the coveralls errors. I know the change is good.

So, the process we run is that when you get a final "r+" comment from me, the next thing you need to do is do one last rebase and squash and re-write the commit message. 

Here it is: r+

Now please rebase and squash it down to just 1 commit. The commit message should be: `fixes 1201143 - consolidating use of STOPWORDS list`

peterbe:
Next, you're now a welcomed contributor to be added to https://air.mozilla.org/contributors/

What you need to do is create a Mozillians profile and get yourself vouched. Once you've set up your profile you can ping me and I'll vouch you in. 

Then, once you've done that (AND if you have a name and a picture on your Mozillian profile) is that you need to apply to be a member of the curated group https://mozillians.org/en-US/group/air-mozilla-contributors/

Ping me when you've applied and I'll take care of adding you [to the list](https://github.com/mozilla/airmozilla/blob/3b530a70b685ce529960694ea4a3cd97cf75dc40/airmozilla/settings/base.py#L312).

64lines:
Thank you @peterbe, I will!

PR #744, opened 2016-08-26
fixes bug 1268717 - misc key/value store for events

peterbe:
@RAMilewski Here's a basic solution (aka. a good start!)
It's a general purpose dirt simple key/value store. Basically, a one step up from a Notes field. 
screen shot 2016-08-26 at 2 09 02 pm

We can do things with this later. Such as a searching by value across all events. Or searching by key.


RAMilewski:
That's brilliant!   It lets Laura and Pearl add anything they want and stop
adding things when the situation changes!!

-- Richard

PR #818, opened 2017-03-25
fixes bug 1350474 - increase font color contrast in footer license note

heatherbooker:
Fixes [bug #1350474](https://bugzilla.mozilla.org/show_bug.cgi?id=1350474) - Change color of font in footer that states license agreement to a lighter color, increasing contrast and readability to meet accessibility guidelines.


heatherbooker:
:) Thank you!

PR #820, opened 2017-04-11
Add pocket employee emails to ALLOWED_BID.

johngian:


peterbe:
This has gone into prod.

PR #824, opened 2018-02-20
redirect sign-in to INXPO

gdestuynder:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1434240 for URL
This is hardcoded as the current decision is to move over to INXPO.


gdestuynder:
closing for now

